PR #220, opened 2026-06-24
v0.52.0: package-level variables (mutable globals)

javimosch:
## What

A top-level **`var name = expr`** is now a **package global**: mutable, shared by every function, type inferred from its initializer and uses. Unlike a local it **persists across calls** — the missing piece for a component to own its state **in machin** rather than in the JS host.

```go
var count = 0
export func bump(d) (n) { count = count + d  n = count }   // accumulates across wasm export calls
```

## Semantics

- **`=`** to a global's name assigns the global; **`:=`** introduces a local (and may shadow the global).
- Visible **everywhere, including closures** — a captured global is referenced directly, not closed over (lambda-lifting already excludes non-local names from capture).
- **Any type:** scalars, strings, **maps** (`var hits = make(map[string]int)` then `hits[k] = …`), **slices** (`var log = []string{}` then `append`).

## Implementation

- **encode:** `splitFunctions` keeps a brace-less top-level `var` as its own declaration.
- **parser:** top-level `var` → `Program.Globals` (parsed after struct registration, so a struct-literal initializer resolves).
- **checker:** globals typed in a synthetic `$globals` context, in declaration order; `*Ident` reads and `=` assignments fall back to a `globalSlot` map (so functions and globals share variables).
- **codegen:** each global is a C `static mfl_g_<name>`; initializers run in a `__attribute__((constructor))` — before `main` (native) and at `_initialize` (wasm reactor). `varRef` renders `mfl_g_<name>` unless a local of the current function shadows it.

## Verified

- Scalar/string/map/slice globals + a closure reading a global run correctly.
- **wasm state persistence:** an exported `bump` accumulates across host calls (3 → 7 → 17), state held in the module.
- New `globals_test.go` (persistence, standalone split, shadowing, codegen shape, wasm target); full suite green, `go vet` clean.

## Version

Bumps the four synced markers to **v0.52.0** (README badge, SPEC, `guide.go`, CHANGELOG); closes gap #5 in the web north star (toward signals + a reactive runtime).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,22 @@
 # Changelog
 
+## v0.52.0
+
+- **Package-level variables — `var name = expr` at top level.** A mutable global
+  shared by every function, its type inferred from the initializer (and its uses).
+  Unlike a local it **persists across calls**, so a wasm module's exported
+  functions can finally hold state between host invocations — `var count = 0` +
+  `export func bump(d) { count = count + d }` accumulates across calls. The piece
+  that was missing for a component to own its state in machin (not the JS host).
+  - `=` to a global's name assigns the global; `:=` still introduces (and may
+    shadow with) a local. Globals are visible everywhere, including inside
+    closures — a captured global is referenced directly, not closed over.
+  - Works for any type: scalars, strings, **maps** (`var hits = make(map[string]int)`,
+    then `hits[k] = …`), and **slices** (`var log = []string{}`, then `append`).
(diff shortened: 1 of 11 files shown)


javimosch:
**Review — Head of Org Engineering:** The diff correctly implements package-level variables as described (CHANGELOG, SPEC, AST, codegen, docs, tests all updated consistently). No syntax errors, out-of-scope edits, security issues, or committed secrets are present.

PR #223, opened 2026-06-25
v0.54.0: reactive computed signals + keyed list reconciliation

javimosch:
## What

Extends `framework/reactive.src` (the signals + patch-list runtime) with the two pieces that make it usable for real components:

### Computed signals
**`computed(func(){ return … })`** — a memoized derived signal, backed by a real signal kept current by a reaction. Read it with `get` and depend on it like any signal. Transitive: a change recomputes the computed, which then updates *its* dependents.

### Keyed list reconciliation
**`each(container, keys, item)`** — `keys()` returns the ordered keys as a CSV string (`func(){ get(ver)  return csv(ids) }`), `item(key)` returns an item's HTML. On a change it emits **only the deltas**: `list_insert` for new keys (rendered once), `list_remove` for gone keys, and a `list_order` directive — **never re-rendering unchanged items**.

The reaction graph is unified: a binding/computed/list recomputes only when a signal it reads changes, and only changed text/keys patch.

**Verified** (native + wasm) on a list+computed app:

| action | patches |
|---|---|
| `add 10` | `total=10, count=1, insert #1, order 1` |
| `add 20` | `total=30, count=2, insert #2, order 1,2`  *(item 1 not re-inserted)* |
| `reverse` | `order 3,2,1`  *(no item re-render)* |
| `remove` | `total=…, remove #k, order …` |

The keys-as-CSV-string design keeps `run_each` type-checkable **even when an app registers no list** (an empty `[]int`-returning closure would default its type and fail).

## Also: builtin shadowing is now a compile error

Defining a function named like a builtin (`flush`, `keys`, `contains`, `len`, …) is now **rejected** — at call sites the builtin wins, so the function was silently ignored (a footgun that bit the reactive runtime *three times*). An `extern` may still shadow a builtin (intentional, for FFI).

## Tests / version

New `reactive_test.go` (minimal-patch behavior end-to-end; shadowing rejected) + `func_slice` tests for the full and no-list runtime. Full suite green, `go vet` clean. Four version markers bumped to **v0.54.0**.

## Summary by CodeRabbit

* **New Features**
  * Added computed values and keyed list updates for reactive content, improving responsiveness and reducing unnecessary re-renders.
  * Introduced a new release version across the changelog, README, and specification.

* **Bug Fixes**
  * Functions can no longer reuse builtin names, preventing confusing naming conflicts; intentional shadowing remains allowed in external interfaces.
  * Improved reactive updates so list changes are applied incrementally, keeping unchanged items intact.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,26 @@
 # Changelog
 
+## v0.54.0
+
+- **`framework/reactive.src` grows computed signals + keyed list reconciliation.**
+  - **`computed(func(){ return … })`** — a memoized derived signal (backed by a
+    real signal, kept current by a reaction); read it with `get` and depend on it
+    like any signal. Transitive: a change recomputes the computed, which updates
+    its own dependents.
+  - **`each(container, keys, item)`** — keyed list reconciliation. `keys()` returns
+    the ordered keys as a CSV string (`func(){ get(ver)  return csv(ids) }`),
+    `item(key)` returns an item's HTML. On a change it emits only the deltas —
+    `list_insert` for new keys (rendered once), `list_remove` for gone keys, and a
+    `list_order` directive — never re-rendering unchanged items. The unified
+    reaction graph means a computed/binding/list all recompute only when a signal
+    they read changes, and only changed text/keys patch.
+- **Defining a function named like a builtin is now a compile error** (it would be
+  silently shadowed by the builtin at call sites — a footgun that bit the reactive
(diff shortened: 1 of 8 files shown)


javimosch:
**Review — Head of Org Engineering:** The diff implements exactly what the PR title/description claims (computed signals, keyed list reconciliation, builtin-shadowing compile error), all changes are syntactically valid and scoped to the reactive framework source, test file, and version/changelog docs, with no correctness bugs, security issues, destructive edits, or leaked secrets.

PR #224, opened 2026-06-25
v0.55.0: reactive templating (slot / list / mount)

javimosch:
## What

`framework/reactive.src` gains a **templating layer** so a component declares its markup **and** its reactive bindings in one place — no hand-written HTML skeleton, no manual `data-s` wiring split across files.

- **`slot(name, compute)`** — returns the markup for a reactive text node (`<span data-s="name">`) and queues its binding.
- **`list(name, keys, item)`** — returns the markup for a keyed-list container (`<div id="name">`) and queues its reconciler.
- **`mount(root, html)`** — sets the root element's HTML once (via a new `dom_mount` host import), then flushes the queued bindings/lists so they paint *after* their elements exist in the DOM.

A whole component becomes one expression:

```go
mount("app",
    "<div class=stats>sum: " + slot("sum", func(){ return str(get(total)) }) + "</div>" +
    list("items", func(){ get(ver)  return csv(ids) }, func(id){ return item(id) }))
```

Static markup is plain string concatenation. The JS host shrinks to a `dom_mount` + the patch ops. **Pure MFL — no compiler change.**

## Verified

- Native + wasm: the component-generated HTML mounts, queued slots paint on mount, the list reconciles, and `add` patches minimally. Robust with **or without** lists.
- `machin-web-demo-reactive` now generates its own markup — `index.html` is just `<div id="app"></div>` + buttons. Same on-screen result (sum/count + sortable items), less hand-wiring.
- New `TestReactiveTemplating`; full suite green; `go vet` clean. Four version markers → **v0.55.0**.

## Why it matters

This is the ergonomics gap before small modular apps (≈30 components, <500 LOC/file) are comfortable — a component is now self-contained MFL (markup + state + reactivity), and the host is generic. The precursor to rewriting the boilerplate as an isomorphic machin app.


javimosch:
**Review — Head of Org Engineering:** The diff does exactly what the PR title/description states — adds slot/list/mount templating to framework/reactive.src, updates the version to 0.55.0 consistently across all version-bearing files (guide.go, README.md, SPEC.md, CHANGELOG.md), and adds a well-scoped regression test. No syntax errors, no stray edits, no secrets, and no correctness bugs are present.

PR #226, opened 2026-06-25
v0.57.0: ptr_str builtin — host→wasm strings (text input)

javimosch:
## What

Adds **`ptr_str(ptr) -> string`** — reads a NUL-terminated string out of raw memory into an MFL (arena) string. This is the missing **into-wasm** string direction.

Strings already flowed *out* of wasm (machin returns a memory pointer the host decodes), and ints flow both ways. Now text can flow *in*: the JS host writes UTF-8 + a NUL into the module's `memory` at a pointer the program `alloc`'d, then calls an export passing that pointer, and the program reads it:

```go
export func input_buf(n) (p) { p = alloc(n + 1) }      // host writes n bytes + NUL here
export func submit(p) { add_todo(ptr_str(p))  free(p) }
```

```js
const bytes = new TextEncoder().encode(inputEl.value);
const p = Number(instance.exports.input_buf(BigInt(bytes.length)));
new Uint8Array(memory.buffer).set(bytes, p);
instance.exports.submit(BigInt(p));
```

**Unblocks forms** — an `<input>`'s text can finally reach a machin component (the gap that kept the isomorphic boilerplate to preset options).

## Verified

- wasm round-trip with multi-byte UTF-8: JS wrote `"héllo from JS 🎉 — user input!"` into wasm memory, machin read it back byte-perfect via `ptr_str`.
- Native poke→`ptr_str` round-trip.
- New `TestPtrStrRoundTrip` / `TestPtrStrTypes`; full suite green, `go vet` clean. Four version markers → **v0.57.0**.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,21 @@
 # Changelog
 
+## v0.57.0
+
+- **`ptr_str(ptr) -> string` — host→wasm strings (text input / forms).** Reads a
+  NUL-terminated string out of raw memory into an MFL (arena) string. This is the
+  missing **into-wasm** string direction: the JS host writes UTF-8 + a NUL into the
+  module's memory at a pointer the program `alloc`'d, then calls an export passing
+  that pointer, and the program reads it with `ptr_str`. (Strings already flowed
+  *out* of wasm as a memory pointer the host decodes; ints flow both ways.) Pairs
+  with `alloc`/`free`. Verified round-trip incl. multi-byte UTF-8 (accents, emoji).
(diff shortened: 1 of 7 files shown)


javimosch:
**Review — Head of Org Engineering:** The change correctly implements `ptr_str` end-to-end: C helper added in codegen.go, type-checker entry in types.go, guide catalog updated, tests cover round-trip and type errors, and version strings bumped consistently across CHANGELOG/README/SPEC/guide.go. No stray edits, no secrets, no correctness or security issues.

PR #227, opened 2026-06-25
v0.58.0: fix calling a captured closure inside a lambda

javimosch:
## What

Fixes a real limitation: **`func(){ fn() }` where `fn` is a captured variable** failed to compile with `call to undefined function "fn"`. The closure was never captured, so by type-check time the lambda had no `fn` in scope.

**Root cause:** closure conversion's free-variable scan (`freeIdents` in `transform.go`) recursed into a `Call`'s **arguments** but ignored its **callee**. So calling a closure-valued variable didn't mark it as free → it wasn't added to the lambda's captures.

**Fix (one line):** in the `*Call` case, mark `ex.Callee` as free when it isn't bound within the lambda. This is safe — top-level function names and builtins aren't bound in the *enclosing* scope either, so the capture filter (`if l.bound[name]`) never picks them up by mistake; only genuine captured variables (params/locals of the enclosing function) are captured.

## Why it matters

It's the higher-order-function building block:

```go
func wrap(fn) (g) { g = func() { return fn() } }        // return a closure calling a captured arg
func g(h) (r) { r = func() { return h(5) } }            // …with an argument
effect(func() { compute() })                            // a stored func() thunk calling a captured compute
```

That last shape is the *clean* reactive model — a single `[]func` of `func()` reactions that capture their own compute closures. The lack of it had forced `framework/reactive.src` into a more verbose typed-per-kind-array design; the runtime can now be simplified (separately).

## Tests

New `closures_test.go` (return-a-closure, closure-with-arg, reactive thunk). Full suite green, `go vet` clean. Four version markers → **v0.58.0**.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed an issue where a captured closure can now be called inside a lambda expression without causing a compile failure.
  * Improved handling so closures used as call targets are recognized correctly.
* **Tests**
  * Added coverage for calling captured closures inside lambdas across multiple cases.
* **Documentation**
  * Updated version information in the changelog, README, specification, and machine-readable guide to `0.58.0`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,17 @@
 # Changelog
 
+## v0.58.0
+
+- **Fix: a captured closure can be CALLED inside a lambda.** `func(){ fn() }` where
+  `fn` is a captured variable now works — previously it failed to compile (`call to
+  undefined function`), because closure conversion's free-variable scan ignored a
+  call's *callee*, so the closure was never captured. This is the higher-order
+  function building block: functions that return closures calling their arguments,
+  and the clean reactive shape `effect(func(){ compute() })` (a stored `func()`
+  thunk that calls a captured compute). Top-level functions/builtins are unaffected
+  (they were never in the enclosing scope, so they're still not captured). One-line
+  fix in `freeIdents`; no regressions.
+
 ## v0.57.0
 
 - **`ptr_str(ptr) -> string` — host→wasm strings (text input / forms).** Reads a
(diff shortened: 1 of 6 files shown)


javimosch:
**Review — Head of Org Engineering:** The change does exactly what the PR title describes: it fixes closure capture of callable variables by adding the callee to `freeIdents` when not already bound. The diff is syntactically valid, scoped to the relevant files (transform.go, a new test, and version bumps), and contains no correctness bugs, security issues, or secrets.

PR #232, opened 2026-06-25
v0.59.0: simplify the reactive runtime + a client-side router

javimosch:
### Fix: closures can capture & mutate an aggregate local
A closure that captures a slice/map/struct local and mutates it failed to compile — the heap box for the captured local was initialized with `*box = {0}`, which is invalid C for an aggregate type (it only compiled for scalars). Now the box is zeroed via `calloc` (the body assigns the real value anyway). So this works:

```go
func acc() (push) { xs := []int{}  push = func(v) { xs = append(xs, v)  return len(xs) } }
// p := acc(); p(10)→1; p(20)→2; p(30)→3
```

### Simplify `framework/reactive.src`
With that fix plus the v0.58.0 captured-closure-**call** fix, the reactive runtime no longer needs its workaround. A **reaction is now just a `func()` thunk that captures its own compute closure** and applies its effect, held in one `[]func` registry:

- `bind(slot, compute)` → a reaction capturing `last` (boxed, persists across runs)
- `each(cont, keyfn, item)` → a reaction capturing `old []int` (the last key set)
- `computed(fn)` → a signal kept current by a reaction

Gone: the typed per-kind arrays (`c_fn`/`c_sig`/`b_fn`/`b_slot`/`b_last`/`e_kf`/`e_if`/`e_cont`/`e_old`) and the `run_computed`/`run_bind`/`run_each` + `r_kind`/`r_sub` dispatch. **~9 fewer globals**, the kind-dispatch removed.

**Same public API**, and **byte-identical patch behavior** — verified `start`/`add`/`sort`/`pop` on the reactive demo produce exactly the same minimal patch list as before (e.g. `sort` → only an `order`, no item re-render).

## Tests / version

New `TestCapturedSliceMutation`; the reactive suite (templating, hydrate, minimal-patches, no-list) still green. `go vet` clean. Four version markers → **v0.59.0**.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,13 @@
   API, **byte-identical patch behavior** (verified), ~9 fewer globals, the dispatch
   gone. (Re-confirmed gotcha: a parameter named like a builtin — `keys` — is
   shadowed at call sites; `each`'s param was renamed `keyfn`.)
+- **`framework/router.src` — a client-side router** (composes with reactive, enabled
+  by the cleaner `reaction()` primitive). The active route is a signal (an int
+  index): `route(path)` registers pages, `navigate(i)` / `nav(path)` switch,
+  `link(path, label)` renders an anchor, and `outlet(id, render)` re-renders the
+  active page (a reaction over a new `dom_html` host import) and syncs the address
+  bar (`nav_url`). `nav` takes the path string from the host via `ptr_str`. For
+  multi-page admin apps / SPAs.
 
 ## v0.58.0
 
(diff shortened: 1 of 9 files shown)


javimosch:
**Review — Head of Org Engineering:** The change does exactly what the PR title/description says: fixes aggregate-type closure capture via calloc, rewrites reactive.src to use a unified func() registry, adds a regression test, and bumps the version string across all four version-bearing files. No syntax errors, security issues, destructive edits, or out-of-scope changes are present.

PR #234, opened 2026-06-25
test: solidify the web domain with framework unit tests

javimosch:
Starts hardening the web domain with MFL-level unit tests for the framework modules.

**machweb** (the server framework) had **zero** MFL-level tests. New `machweb_test.go` composes `framework/machweb.src` with a test app and runs the pure request/response/router logic natively (the net builtins compile but `serve()` is never called):
- `parse_request` — method, path, body, and a lowercased header map; `header()` case-insensitive + `""` when absent.
- `param` (segment after a prefix), `content_length` (incl. absent → 0).
- response builders — status line, content type, `is_bin=0` for text.
- map router `dispatch` — method-aware key, 404 fallback for unknown route **and** wrong method.

**reactive** — pin the load-bearing guarantees that the runtime simplification (v0.59.0) must preserve:
- change detection — `set(id, equalValue)` emits no patch.
- dependency isolation — only reactions that read a signal re-run.
- computed transitivity — `set(base)` updates a computed → patches its slot.
- minimal `each()` deltas — a removal is one `list_remove` (no re-insert); a pure reorder is `list_order` only.

**router** — `path_index` resolution (and default-to-0 for an unknown path) + navigate-by-path.

Tests only; no compiler or framework changes. Full suite green.

### DIFF
--- a/machweb_test.go
+++ b/machweb_test.go
@@ -1,0 +1,145 @@
+package main
+
+import (
+	"os"
+	"strings"
+	"testing"
+)
+
+// machwebProg composes framework/machweb.src (the server framework) with a test
+// app and parses it. machweb has no extern block — its net builtins (listen/accept/
+// read/write) compile but are only reached from serve(), which the tests never call,
+// so the pure request/response/router logic runs natively under RunCaptured.
+func machwebProg(t *testing.T, app string) *Program {
+	t.Helper()
+	data, err := os.ReadFile("framework/machweb.src")
+	if err != nil {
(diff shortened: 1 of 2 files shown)


javimosch:
**Review — Head of Org Engineering:** The diff adds only new test files/functions (machweb_test.go is new, reactive_test.go gets new test cases appended), is syntactically valid Go, and contains no secrets, destructive changes, or correctness bugs. It does exactly what a PR adding framework tests would be expected to do.

PR #235, opened 2026-06-25
test: SQLite framework tests + make cover (78.2%)

javimosch:
Continues hardening the web domain and answers the "what's our coverage?" question with a repeatable number.

**SQLite** (`sqlite_test.go`) — the data layer behind the CRUD back-office, now with native round-trip tests (`build.go` auto-links `-lsqlite3`):
- open `:memory:` → exec → query round-trip; the result set is a JSON array that `json_get` indexes by row/field.
- **`?`-param binding is injection-safe** — a `'; DROP TABLE ... `value is stored as data (the table survives with both rows), and a value with quotes round-trips verbatim. Proof it was bound, not interpolated.
- UPDATE / DELETE are observable on the next query (the write path the back-office reloads from).

**Coverage** — `make cover` runs the suite with `-coverprofile`, prints statement coverage, and writes `coverage.html` (highlighting unhit lines). The corpus is the MFL programs the tests compile+run, so the number tracks how much of the **toolchain** the suite exercises. **Currently 78.2%.**

> Note on what the % measures: it's Go-statement coverage of the *compiler*. The SQLite C runtime is emitted C (not Go), so these tests don't move the % even though they're real behavioral coverage — they exercise the codegen/typecheck paths (already hit) plus the runtime end-to-end. A future MFL-native coverage tool would measure framework `.src` line coverage; see the PR discussion.

Tests + Makefile + .gitignore only.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,7 @@
 /machin
 /bin/
 /framework/app.mfl
+
+# coverage reports (make cover)
+coverage.out
+coverage.html
--- a/Makefile
+++ b/Makefile
@@ -4,7 +4,7 @@ BIN     := bin/machin
 PREFIX  ?= /usr/local
 GOFLAGS ?= -trimpath
 
-.PHONY: all build test examples bench install uninstall clean
+.PHONY: all build test cover examples bench install uninstall clean
 
 all: build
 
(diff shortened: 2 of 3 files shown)


javimosch:
**Review — Head of Org Engineering:** The diff adds coverage tooling (Makefile target + .gitignore entries) and new SQLite integration tests that match the PR's stated scope. The code is syntactically valid, touches only appropriate files, and contains no secrets, destructive changes, or correctness bugs.

PR #237, opened 2026-06-25
test: machweb socket I/O path (multi-segment body + binary response)

javimosch:
Closes the last untested machweb seam — the real socket I/O path (`read_request`'s multi-read loop and the `write_bytes` binary path), which the pure request/response tests in #234 couldn't reach.

Each test is a **one-shot in-process loopback** under `RunCaptured`: `main` listens, a goroutine accepts + runs `machweb_handle` (read → parse → handler → write → close), and `main` itself is the HTTP client (`dial` → `write` → `read_all`). So it exercises the actual `listen`/`accept`/`read`/`write`/`write_bytes`/`close` builtins end-to-end, not just the parsing functions.

- **Multi-segment body** — the client sends headers + the first 5 bytes of an 11-byte body, pauses, then sends the rest. `read_request` must loop until Content-Length bytes arrive; proven by the handler echoing the **full** `Hello World`, not just the first segment.
- **Binary response (NUL-safe)** — handler returns `ok_bytes` whose body starts with `0x00`. The binary branch writes `write_bytes(conn, bin)` and reports `Content-Length: 4`; a text/strlen path would have said `0`. That length in the response head is the proof the bytes path ran.

Stable across `-count=3` (`SO_REUSEADDR` is set, so quick re-runs don't `EADDRINUSE`). Tests only.

### DIFF
--- a/machweb_io_test.go
+++ b/machweb_io_test.go
@@ -1,0 +1,100 @@
+package main
+
+import (
+	"strings"
+	"testing"
+)
+
+// Integration tests for machweb's socket I/O path — the part the pure-logic tests
+// can't reach. Each program is a one-shot loopback: main listens, a goroutine
+// accepts + runs machweb_handle (read_request -> parse -> handler -> write), and
+// main itself is the HTTP client (dial -> write -> read). This exercises the real
+// read/write/close builtins, not just the parsing functions.
+//
+// read_all drains the socket until the server closes (read returns "").
+const loopbackHelpers = `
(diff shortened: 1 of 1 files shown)


javimosch:
**Review — Head of Org Engineering:** The diff adds a new test file (`machweb_io_test.go`) that exercises the socket I/O path as described, is syntactically valid Go, touches only the new file, and contains no secrets, destructive changes, or correctness bugs.

PR #239, opened 2026-06-25
v0.60.0 — pure-MFL PostgreSQL client (SCRAM-SHA-256)

javimosch:
The first **networked datastore** for machin, and the start of the SME-backend dogfood (`docs/NORTH-STAR-BACKEND.md`). You picked the heaviest option — here it is, working end to end against `postgres:16`.

```go
type User struct { id int  name string  email string }
pg_connect("127.0.0.1", 5432, "postgres", "machindb", "secret")
rows  := pg_query("SELECT id, name, email FROM users ORDER BY id")
users := parse(rows, []User{})        // same shape as sqlite_query — typed decode
pg_disconnect()
```

### What's here
- **`framework/postgres.src`** — a PostgreSQL client speaking **wire protocol v3** over `dial()` with **SCRAM-SHA-256** auth (the modern PG default), in **pure MFL — no libpq, no cgo**. `pg_query` returns a **JSON-array-of-rows string in the same shape as `sqlite_query`**, so `parse(rows, []T{})` decodes it; numeric/bool columns come back unquoted (typed via the result's column OIDs), `NULL` → `null`, strings JSON-escaped. v1 is the simple-query protocol.

### Two language gaps it surfaced (dogfood → filled)
- **`read_bytes(fd) -> bytes`** — NUL-safe socket read (`read()` is a C string and truncates at the first 0 byte; a binary wire protocol can't use it).
- **`base64_encode_bytes` / `base64_decode_bytes`** — binary-safe base64 (the string forms stop at a NUL), for SCRAM salts/proofs.
- (SCRAM's XOR folds use MFL's native bitwise operators — no builtin needed.)

### Verification
- `wire_test.go` (**CI-runnable**): binary base64 round-trips through NULs + matches the string form on ASCII; `read_bytes` loopback returns a leading-NUL buffer intact.
- `postgres_test.go` (**gated by `MACHIN_PG_TEST`** so CI never blocks on a missing DB): SCRAM auth + query, asserting typed decode, a quote round-trip, and bool/NULL handling. Verified locally:
```
rows=[{"id":1,"name":"Ada Lovelace","email":"ada@example.com"},{"id":2,...}]
escaped+null: [{"id":3,"name":"O'Brien \\ \"x\"","email":null}]
count+bool: [{"n":3,"flag":true}]
```
Full suite + `go vet` green; `machin guide` version/typecheck test passes.

### Roadmap (in the north-star doc)
1. **Postgres v2 — parameterized queries** (extended protocol; the key follow-up, v1 simple-query is for trusted SQL).
2. Sessions/cookies in machweb → 3. SSO library → 4. Redis client.

Bumped to **v0.60.0** (new builtins + framework).

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -47,6 +47,13 @@ verify, and the gotchas. A runnable data-layer reference is
 [`examples/complex/sqlite_crud.mfl`](examples/complex/sqlite_crud.mfl); direction
 and gap roadmap live in [`docs/NORTH-STAR-WEB.md`](docs/NORTH-STAR-WEB.md).
 
+Building an **SME backend** (a datastore client, auth, a service integration)? The
+direction and capability/gap matrix live in
+[`docs/NORTH-STAR-BACKEND.md`](docs/NORTH-STAR-BACKEND.md). machin already has embedded
+SQLite and a pure-MFL **PostgreSQL** client ([`framework/postgres.src`](framework/postgres.src)
+— wire protocol + SCRAM-SHA-256, no libpq); both return rows as JSON that
+`parse(rows, []T{})` decodes.
+
 Building a **game** (terminal TUI or raylib GUI/audio)? Read
 [`skills/machin-gamedev/SKILL.md`](skills/machin-gamedev/SKILL.md) first — the
 canonical setup, build-and-verify workflow, raylib FFI surface, and accumulated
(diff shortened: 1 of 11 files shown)


javimosch:
**Review — Head of Org Engineering:** The diff does what the PR title/description says (adds v0.60.0 PostgreSQL client, binary-safe base64, and read_bytes primitives), all changed files are in scope (codegen.go, docs, changelog, readme, spec, agents.md), the C implementations are syntactically valid and algorithmically correct, and no secrets or destructive changes are present.

PR #263, opened 2026-06-29
bench: native-speed suite — machin vs Rust vs Zig

javimosch:
## Why

The README claimed "C/Rust-class speed" on the strength of a single `fib(40)` line. This backs it with a real four-kernel suite against the two reference native toolchains.

## What

`bench/native-speed/` — `fib`, `mandelbrot`, `sieve`, `intsum`, each written in **machin / Rust / Zig**, each producing **byte-for-byte identical output** (verified every run, so we time the *same* computation). Each language at its standard release setting: Rust `-C opt-level=3` (its real release, not the weaker `-O`), Zig `-OReleaseFast`, machin `cc -O2`. No `-march=native` for anyone.

## Result (this machine; min of 5)

| kernel | machin | Rust -O3 | Zig fast | machin vs best |
|---|--:|--:|--:|--:|
| fib(40) | **245 ms** | 303 | 306 | 1.00× |
| mandel 1000² | 827 | **814** | 819 | 1.02× |
| sieve 10⁷ | 203 | 153 | **145** | 1.40× |
| intsum 10⁹ | **2832 ms** | 3764 | 3556 | 1.00× |

**machin wins 2, ties 1, loses 1.** Because it *is* C underneath: gcc -O2 on the generated C matches or beats rustc -O3 / Zig on scalar recursion + integer loops here, ties on float, and trails ~1.4× on array-heavy indexing (a real, documented limitation — not hidden).

The point isn't "machin beats Rust" — it's that machin sits squarely in the **compiled-native tier**, reachable from source an agent writes about as cheaply as Python (the [`rest-sqlite`](../rest-sqlite) bench).

Docs/bench only — no compiler change, no version bump.

### DIFF
--- a/README.md
+++ b/README.md
@@ -98,7 +98,18 @@ Built something? Add it to [awesome-machin](https://github.com/javimosch/awesome
 
 ## Performance
 
-`fib(40)` (~331M calls): **MFL 0.20s** · C 0.19s · Rust 0.29s. Values are unboxed; no interpreter, no VM.
+machin compiles through C, so it runs in the **native tier**, not the scripting tier.
+On four kernels with byte-identical output, `cc -O2` on machin's generated C **wins 2,
+ties 1, loses 1** against Rust `-O3` and Zig `ReleaseFast` (min of 5, this machine):
+
+| | fib(40) | mandelbrot 1000² | sieve 10⁷ | intsum 10⁹ |
+|---|--:|--:|--:|--:|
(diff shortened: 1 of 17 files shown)


javimosch:
**Review — Head of Org Engineering:** The PR adds a benchmark suite under bench/native-speed/ and updates the README performance section to reflect the new results — exactly what the title describes. The diff is scoped, syntactically valid, contains no secrets or destructive changes, and touches only new files plus the README performance blurb.

PR #265, opened 2026-06-29
skill: machin-start — the agent-facing decision + bootstrap entry point

javimosch:
## Why

The benchmarks are ammunition; this is the thing that fires them. machin's four existing skills are domain how-tos whose `description:` fields all say *"Use when writing or debugging a machin X"* — they only fire **after** machin is chosen. Nothing fired at the **decision moment** (a dev wants a small deployable service, stack still open). That moment is where adoption begins, and it was empty.

## What

A fifth embedded skill, **`machin-start`**, sitting above the domain four:

1. **When to reach for machin** — a decision table backed by the three `bench/` results (agent-write cost ties Python; native-tier speed; 92.9 kB / 0.49 ms / 0.1 MB ship). Plus an **honest when-NOT-to** (existing ecosystems, needs a mature library, data-science, team won't run a young language).
2. **Zero → running → shipped quickstart** — install → a complete REST+SQLite service → `machin build` → the `CC=./muslcc` static-musl step → a `FROM scratch` image. (Snippet verified to compile here.)
3. **Routes** to `--skill web|backend|gamedev|deploy`.

Wired exactly like the existing skills: `//go:embed` + `machin guide --skill start` (alias `machin`) + listed **first** in the `machin guide` domain registry.

## The distribution lever

A skill that only lives in this repo fires only when an agent is *already in* this repo — useless for reaching new users. So `install.sh` now also registers the skill into a **vendor-neutral `~/.agents/skills/machin-start/SKILL.md`** (not Claude-specific). The binary emits its **own embedded skill** (`machin guide --skill start > …`), so it needs no extra download and never drifts. Opt out with `MACHIN_NO_SKILL=1`; relocate with `AGENT_SKILLS_DIR`. Source of truth stays in machin.

## Scope notes

- **Docs-only** — no `machin new` scaffold yet (deferred), no compiler change, **no version bump**. Added to CHANGELOG Unreleased.
- Guide tests pass; quickstart compiles to the same 44 kB binary as the bench app.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,15 @@
 
 ## Unreleased
 
+- **New embedded skill `machin-start`** (`machin guide --skill start`) — the
+  agent-facing entry point that comes *before* the domain how-tos: when to reach
+  for machin (and when not), with the measured `bench/` numbers as decision
+  criteria, plus a zero→running→shipped quickstart. `install.sh` now also registers
+  it into a vendor-neutral `~/.agents/skills/machin-start/SKILL.md` (the binary
+  emits its own embedded skill; opt out with `MACHIN_NO_SKILL=1`, relocate with
+  `AGENT_SKILLS_DIR`), so a coding agent surfaces machin at the decision moment.
+  Listed first in the `machin guide` domain registry.
+
 Driven by **machin-wiki** (local) — the first wasm client that *initiates* server calls
 through returning `extern "env"` imports (`data := http_get(url)`, used inline), with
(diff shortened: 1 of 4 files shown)


javimosch:
**Review — Head of Org Engineering:** The PR does exactly what it describes: adds a new embedded `machin-start` skill (valid Markdown, valid Go embed directive, correct switch case and error message update, safe shell script with opt-out). No syntax errors, no out-of-scope edits, no secrets, no destructive changes, and no no-touch areas are touched.

PR #266, opened 2026-06-29
skill: `machin skill install` — close the skill-discovery loop

javimosch:
## Why

`machin-start` (the agent-facing decision skill, #265) only helps if an agent can *see* it. A skill living in this repo only fires when an agent is already in this repo — which is no one we're trying to reach. The old `install.sh` path wrote it to `~/.agents/skills`, but **only on a curl|sh install** (source builds got nothing) and **only to a directory no runtime reads**. The loop was open.

## What

**`machin skill install`** — writes all five embedded `SKILL.md` files where coding agents actually look:
- the vendor-neutral `~/.agents/skills/` (canonical), **plus**
- any **detected runtime** — Claude Code's `~/.claude/skills/` (the same `SKILL.md` format these already use), targeted only if it exists so we never litter.

Works **regardless of how machin was installed** (curl|sh *or* `make build`) — the gap that left this machine with zero registered skills despite machin being installed. Rounded out with `machin skill list`, `machin skill show <name>`, and `--dir <path>` (for a project's `.claude/skills`). `install.sh` now just calls it; `MACHIN_NO_SKILL=1` opts out.

**Verified live:** ran `machin skill install` on this machine → the 5 skills registered into `~/.claude/skills/` and the session's tooling immediately surfaced `machin-start` + the four domain skills as available. Loop closed, demonstrably.

## Also

- Rename `bench/cold-start/Dockerfile.go` → `go.Dockerfile` (+ siblings) so `go vet ./...` / `go test ./...` stop trying to parse a Dockerfile as Go. `go vet ./...` is now clean.

No version bump (Unreleased). Full `go test .` green.

## Summary by CodeRabbit

* **New Features**
  * Added a new `machin skill` command with `list`, `show`, and `install` options.
  * Users can now install built-in skills to detected skill directories, with support for custom install paths and installing all skills at once.
  * Startup installation now uses the built-in skill installer and can be skipped with `MACHIN_NO_SKILL=1`.

* **Bug Fixes**
  * Fixed cold-start benchmark files so Go tooling no longer treats them as Go source during checks and tests.


javimosch:
**Review — Head of Org Engineering:** The diff does exactly what the PR title/description says: renames Dockerfile.* to *.Dockerfile (fixing go vet false positives), updates images.sh to match, and adds the `machin skill install` command in a new well-scoped skillcmd.go. No syntax errors, no secrets, no destructive changes, and no out-of-scope edits.

Issue #260, opened 2026-06-26
Server-side TLS / STARTTLS: a wrap-an-fd-in-TLS primitive (native HTTPS + SMTP STARTTLS)


javimosch:
## Gap: server-side TLS / STARTTLS — a "wrap an existing fd in TLS" primitive

machin has a TLS **client** (`https_get`/`https_post`, the `wss://` client) where dial + handshake are fused. What's missing is the ability to **wrap an already-connected socket fd in TLS**, which unlocks two things at once:

1. **Native HTTPS termination in machweb** — `serve_tls(port, cert, key, handler)` so a single machin binary can face the internet without a reverse proxy (the no-proxy / edge / internal-service case; eventually the ACME "auto-HTTPS" story).
2. **STARTTLS for line protocols** — SMTP (submission to Gmail/SendGrid/SES, which require TLS), IMAP, FTPS… all do `dial plaintext → negotiate → upgrade the same fd to TLS`.

### Why now
Surfaced by two dogfood apps this cycle:
- **machin-mail** (v0.80.0, `framework/smtp.src`) — plaintext SMTP + AUTH works, but a public relay requires STARTTLS, which needs this primitive.
- **machin-deploy** (v0.78.0) — proxy-aware hardening shipped; native TLS termination was deliberately deferred because a reverse proxy covers HTTPS today. This primitive is what would make machweb optionally self-terminating.

### Sketch
A server-side + upgrade TLS surface, e.g.:
- `tls_server_ctx(certfile, keyfile) -> ctx`
- `tls_accept(ctx, fd) -> tls`  (server handshake on an accepted socket)
- `tls_client_fd(fd, hostname) -> tls`  (client handshake on a connected socket — STARTTLS)
- `tls_read(tls) -> bytes` / `tls_write(tls, bytes) -> int` / `tls_close(tls)`

The catch: machin's `read`/`write`/`read_bytes`/`write_bytes` operate on raw fds, so TLS connections need a parallel SSL-aware read/write path (the `wss` client already has SSL read/write internally — factor it out and expose it on an opaque `tls` handle). OpenSSL is already linked.

### Dogfood to validate (zero-dep where possible)
- `serve_tls` → drive with `curl --insecure` against a self-signed cert.
- STARTTLS → machin-mail `send` to a TLS-capable local catcher (would need the sink to also gain `tls_accept`, closing the loop).

Labels: enhancement, networking, tls

javimosch:
[...]

PR #268, opened 2026-06-29
build: `machin build --static` — bundle SQLite, ship FROM scratch

javimosch:
## The gap (surfaced twice)

Both the flagship (machin-hook) and the adoption-loop validation hit the same wall: **machin had no turnkey static build for SQLite apps** — the most common backend shape. `machin build` links system `libsqlite3`, so a `FROM scratch` binary needed the SQLite amalgamation compiled in *by hand*. The validation agent only pulled it off because a copy of `sqlite3.c` happened to be cached on the box; a stranger would've been stuck.

## The fix: `machin build --static`

- The public-domain **SQLite amalgamation v3.53.3** is vendored gzipped and `//go:embed`'d into the compiler (~2.6 MB), and **compiled directly into the program** in `--static` mode — so a SQLite app links **no `libsqlite3`**.
- Paired with `CC=musl-gcc`, you get a **libc-free ~1 MB binary that runs `FROM scratch`**.
- `BuildBinary` is now a thin wrapper over `BuildBinaryStatic` — all existing callers/tests untouched.

## Verified

| build | result |
|---|---|
| `CC=musl-gcc machin build --static` (rest-sqlite) | 1.23 MB, `statically linked`, **runs from a `FROM scratch` image** serving SQLite-backed traffic |
| `machin build --static` (plain glibc cc) | 2.26 MB, statically linked, SQLite works — so the regression test needs no musl |
| `--static` on a TLS app (machin-hook) | prints an honest heads-up, then fails on `openssl/ssl.h` (as documented) |

New `static_test.go` asserts a `--static` SQLite binary doesn't link `libsqlite3` and still runs (`"x":7`).

## Scope (as agreed)

- **In:** SQLite, embedded (turnkey/offline/deterministic — exactly the fragility the validation exposed).
- **Out:** TLS/crypto (OpenSSL) — no single-file amalgamation; the right fix is **native TLS (#260)**, which removes the OpenSSL dependency entirely. `--static` warns clearly on TLS/crypto programs.

`machin-start` + `machin-deploy` skills updated so the `FROM scratch` path is now a **turnkey command**, not a caveat. No version bump (Unreleased — there's now a solid batch for a 0.81.0 whenever you want to cut it). Full `go test .` green, `go vet ./...` clean.


javimosch:
**Review — Head of Org Engineering:** The change implements exactly what the PR title describes (machin build --static with bundled SQLite amalgamation), the Go and Markdown edits are syntactically valid, it touches only relevant build/main/skill files with no stray edits, and there are no correctness bugs, security issues, or leaked secrets — the static linking logic, temp-dir cleanup via defer, and conditional bundling are all sound.

PR #282, opened 2026-07-02
machin guide proof section — reproducible credibility, agent-legible

javimosch:
## Summary
- Adds a `proof` section to `machin guide`'s JSON catalog (schema bumped `v2` -> `v3`): `selfHosting` (the bootstrap-fixpoint claim) and `benchmarks` (the three measured comparisons — agent write cost vs Go/Python, native runtime speed vs Rust/Zig, cold-start/ship-size vs Node — that already live in README/bench/), each with a `reproduce` command pointing at the real `bench/` script.
- This is the "next big move" landed the agent-first way: instead of a human-facing writeup (LinkedIn/founder-narrative, which was the first idea and got explicitly overridden — machin's positioning is machine-first, not human-DX), the credibility mechanism is an agent running `machin guide` while researching what to build with, and relaying the numbers to its principal.
- `--text` gets a PROOF section rendered alongside COMMANDS/DOMAINS/TYPES/etc.
- New `TestGuideProofReproducible` asserts every `reproduce` path is a real directory in the repo, matching how `TestGuideBuiltinsRecognized`/`TestGuideIdiomsCompile` already keep the rest of the catalog honest — "reproducible, not asserted" stays true as the repo evolves.

## Test plan
- [x] `go build ./...` clean
- [x] `go test .` full suite passes (incl. new `TestGuideProofReproducible`)
- [x] `machin guide` JSON and `machin guide --text` both render the new section correctly
- [x] `TestGuideCatalog` extended to assert the proof section is populated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,18 @@
 
 ## Unreleased
 
+- **`machin guide` gained a `proof` section — the "why trust these claims" answer, as
+  structured data an agent can read and relay, not a blog post a human has to find.**
+  Schema bumped to `machin.guide/v3`. `proof.selfHosting` states the self-hosting fixpoint;
+  `proof.benchmarks` carries the three measured comparisons (agent write cost vs Go/Python,
+  native runtime speed vs Rust/Zig, cold-start/ship-size vs Node) that already lived in
+  README/bench/, each with a `reproduce` command pointing at the actual `bench/` script — not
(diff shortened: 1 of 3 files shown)


javimosch:
**Review — Head of Org Engineering:** The PR adds a `proof` section to the guide catalog as described, bumps the schema version to v3, renders it in `--text` output, and adds `TestGuideProofReproducible` to keep reproduce paths honest — all within scope and structurally valid Go. No correctness bugs, security issues, destructive changes, or out-of-scope edits are present.

PR #284, opened 2026-07-02
Static TLS builds, FROM scratch (embedded CA bundle) — closes #283

### LINKED ISSUE #283: Static TLS/crypto binaries: bundle a CA store — the FROM-scratch gap is smaller than assumed
## Gap: --static + TLS already links, it just can't verify a cert FROM scratch

Distinct from #260 (server-side TLS/STARTTLS — a *capability* gap; OpenSSL is already
dynamically linked there). This is a *packaging* gap: `machin build --static` on a
program using `https_get`/`http_request`/crypto builtins already attempts `-static
-lssl -lcrypto` (`build.go:108-113`) but only warns "the link needs static OpenSSL
libs present, or it will fail" — implying this needs real engineering (vendoring a
TLS stack, months of work). **It does not. It already works, with one missing piece.**

### What I verified (spiked, not speculated)

- This host's `libssl-dev` already ships `libssl.a`/`libcrypto.a` (glibc, standard
  `apt install libssl-dev` — **no musl, no vendoring needed for the crypto itself**,
  unlike the SQLite `--static` story which specifically wants `musl-gcc`).
- `machin build --static` on an `http_get("https://...")` program **already
  compiles and links successfully** against those archives, right now, no code
  changes. Stripped binary: ~5.0 MB (not 92.9 kB — a materially different, but
  still real, number; see "Sizing honesty" below).
- Ran it for real (not just `ldd`): on the host, actual DNS + TLS handshake +
  HTTP 200 — works.
- **The actual acid test**: copied the stripped binary into a genuine `FROM
  scratch` Docker image (nothing else — no `/etc`, no shared libs) and ran it.
  **Fails**: `ERR: tls`. Diagnosis confirmed by elimination: `SSL_CTX_set_default_verify_paths`
  (`codegen.go:1267`, in `mfl_ssl_ctx()`) expects `/usr/lib/ssl/certs` to be a
  **directory of hash-named symlinks** (the traditional `c_rehash` layout), not a
  flat CA bundle file — so a scratch image with no such directory structure has
  no way to verify any server certificate. DNS/connect were NOT the problem
  (`err == "tls"` means it got past both).
- **Confirmed the fix**: re-ran the same scratch image with `SSL_CERT_FILE=/path/to/single-file-bundle.pem`
  set (pointing at one flat PEM bundle, which OpenSSL *does* accept in single-file
  form) — works, HTTP 200, real cert verification against a real CA.

### The bounded v1

1. Vendor a public CA root bundle (Mozilla's, the same one curl ships — public,
   small, ~200 KB) gzipped and `//go:embed`'d, same pattern as
   `vendor/sqlite/sqlite3.c.gz` in `build.go`.
2. In `mfl_ssl_ctx()` (`codegen.go:1261-1271`), when compiled for a static build
   (a new `-DMFL_STATIC_CABUNDLE` / a compiled-in byte array alongside the
   generated C, same mechanism as the SQLite amalgamation source file), load the
   embedded bundle into the `SSL_CTX`'s `X509_STORE` from memory (`BIO_new_mem_buf`
   + repeated `PEM_read_bio_X509` + `X509_STORE_add_cert`) as a fallback *alongside*
   `SSL_CTX_set_default_verify_paths` — dynamic (non-static) builds keep trusting
   the live system CA store (correct — it gets security updates); only static
   builds fall back to the embedded snapshot when the system store isn't there.
3. Update `build.go`'s static-TLS warning (currently implies "this will probably
   fail") to reflect reality: it links fine today; document that a truly
   dependency-free `FROM scratch` run needs the embedded-bundle piece (item 2) to
   verify certs, and that static TLS does NOT require `musl-gcc` (glibc's own
   static archives are enough) — simplifying, not complicating, the existing docs.
4. A Go test building a `--static` TLS program and running it in a genuinely
   isolated environment (mirroring the Docker spike above), plus a negative case
   (a self-signed/untrusted cert must still be rejected — proving the embedded
   bundle does real verification, not `SSL_VERIFY_NONE`).
5. A new `bench/` entry (or a row in the existing `cold-start` one) with the real,
   honest number — a static TLS-capable binary, not the libc-only 92.9 kB figure —
   feeding `machin guide`'s `proof.benchmarks` (shipped in v0.89.0) as a 4th entry.

### Sizing honesty (don't let this become a new false claim)

The static TLS/crypto binary is **~5 MB stripped**, not 92.9 kB — OpenSSL is a big
static archive. That's still 35x smaller than Go's own static binary story for
comparable apps and ~2-3 orders of magnitude below Node/Python containers, but it's
a **different, separate bench row** from the existing libc-only figure — don't
merge the two numbers or the `proof` section (which this repo just spent real
effort making "reproducible, not asserted") becomes the thing it was built to avoid.

### Why now

Surfaced originally by machin-hook's "honest deploy finding" (README/CHANGELOG,
v0.81/v0.82 era): apps using `http_request`/sqlite together can't ship truly
dependency-free. The SQLite half of that gap is closed (`--static`, v0.81.0). This
closes the TLS half — and turns out to be a CA-bundle problem, not a TLS-stack
problem.

Labels: enhancement, tls, static-build

javimosch:
## Summary
- `machin build --static` on a TLS/crypto-using program already linked OpenSSL statically (this host's `libssl-dev` ships static `.a` archives — no vendoring, no musl needed for that part). The only missing piece was certificate verification with no system CA store, so it failed in a genuinely empty `FROM scratch` container.
- Vendors Mozilla's CA root store (as shipped by Debian/Ubuntu's `ca-certificates`, ~245 KB gzipped) the same way the SQLite amalgamation is vendored: gzipped, `go:embed`'d, compiled directly into static builds as a byte array, loaded into the `SSL_CTX` as a fallback alongside the system store.
- **Verified end-to-end in a real Docker `FROM scratch` image** (nothing else in it — no `/etc`, no shared libs): a genuine HTTPS request with full certificate verification succeeds, and a known self-signed/untrusted cert is correctly **rejected** — proving verification is actually active, not silently disabled.
- New `bench/tls-static` (26.5 kB dynamic vs 5.28 MB static) feeds a 4th `proof.benchmarks` entry in `machin guide` — kept as its own honest number, not folded into the existing 92.9 kB libc-only figure.
- Distinct from issue #260 (server-side TLS/STARTTLS — a separate, still-open capability gap). Closes #283.
- Static TLS pairs with the **default `cc`**, not `CC=musl-gcc` (the static OpenSSL archives here are glibc-built) — docs (`skills/machin-start/SKILL.md`, synced to the plugin copy) updated accordingly.

## Test plan
- [x] `go build ./...` clean
- [x] `go test .` full suite passes, incl. new `TestStaticBuildBundlesCACerts` (structural: no dynamic libssl/libcrypto, bundle really compiled in; behavioral: untrusted local cert rejected — no live-internet dependency in CI)
- [x] Manually verified in Docker: positive case (real HTTPS request, real cert, `FROM scratch`, zero files) and negative case (`self-signed.badssl.com` correctly rejected)
- [x] `bench/tls-static/run.sh` reproduces the measurement end-to-end, including the Docker FROM-scratch check when Docker is available
- [x] `machin guide` / `--text` render the new benchmark row correctly; `TestGuideProofReproducible` passes


javimosch:
**Review — Head of Org Engineering:** The change does what the PR title/description says (embeds a CA bundle so --static TLS builds work FROM scratch), all added code and files are syntactically valid, no stray edits or out-of-scope changes, no secrets or destructive operations, and no correctness/security defects are present.

PR #285, opened 2026-07-02
feat: inferred data-race freedom — Rust's guarantee, no Send/Sync (go/chan/select safety)

javimosch:
Adds an **inferred data-race checker** to `machin check` — the guarantee Rust gives via `Send`/`Sync`, with **zero annotations**. machin already had `go`/channels/`select`; this makes concurrent MFL provably race-free (or tells you why, with a counterexample). Surfaced as errors in `check`/`--json` (phase `race`) and gated behind `machin build|run --race-safe`; plain `build` is unaffected.

## What it covers (every way MFL shares mutable state across threads)
- **Locals across goroutines** — type-aware (reachability: a slice-field of a by-value struct still shares its backing; a scalar field doesn't), reads + writes, loop-spawn multiplicity. `RACE001`/`RACE002`.
- **Package globals** — one shared cell, so even a scalar global races. `RACE001`/`RACE002`.
- **Channel ownership** — move-on-send: using a value after `ch <- v` is `RACE004` use-after-move.
- **Closure captures** — a capturing closure passed to a `go`-spawned function that invokes it (the only way a closure reaches a goroutine in MFL).
- **Happens-before precision (sound)** — accesses before a spawn, or after a channel-join barrier, are ordered and not flagged. Only suppresses when provable.

## Soundness
Borrow-checker discipline: over-approximate, never a false negative on the covered surface. Adversarially tested (signal-before-write, too-few-receives, post-spawn all still flag). Validated against the 5 real concurrency apps (healthcheck/linkcheck/pipe/pool/wscat): **0 false positives**; a mutation test confirms it engages. Corpus validation even caught a soundness bug in the first cut (loop multiplicity), now fixed + regression-tested.

Docs: `docs/check-json.md` (the RACE taxonomy) + `machin guide` (agents discover it) + `docs/concurrency-race-freedom.md` (the full design). New `racecheck.go` + `racecheck_test.go`; full `go test .` green.


javimosch:
**Review — Head of Org Engineering:** The change correctly wires the race-detection phase into analyzeSource by capturing the Check result and calling detectRaces on a clean typecheck, with diagnostics mapped back to source lines — this is syntactically valid Go and does exactly what the PR title describes. The doc updates (check-json.md phase enumeration and code table, new blog draft) are additive and accurate; no stray edits, secrets, or destructive changes are present.

PR #286, opened 2026-07-02
Server-side TLS + STARTTLS — machweb terminates HTTPS itself

javimosch:
## Summary
- New builtins: `tls_server_ctx(certfile, keyfile) -> int`, `tls_accept(ctx, fd) -> int` (server handshake), `tls_client_fd(fd, hostname) -> int` (the STARTTLS primitive — upgrade an already-connected, plaintext-negotiated fd to verified TLS in place), and `tls_read[_bytes]`/`tls_write[_bytes]`/`tls_close` mirroring the plain fd builtins over a tls handle.
- Turned out `serve()`/`machweb_handle()` (`framework/machweb.src`) are pure MFL — no Go-side server plumbing to touch. So this didn't need to touch machweb's routing/response logic at all: new `serve_tls(port, certfile, keyfile, handler)` is `serve`'s TLS-terminating sibling, reusing the same router/handler/`Response` machinery via a deliberate near-duplicate of `machweb_handle` rather than threading a protocol-polymorphic abstraction through every call site.
- **v1 scope** (documented in the guide's `server-tls-v1` gotcha): one cert per `tls_server_ctx` (no SNI multi-cert), no client-cert verification (not mutual TLS), no ACME/auto-renewal, and `res.is_hijack`/`res.is_stream` get a 501 over `serve_tls` rather than misbehaving — a deliberate, documented limitation, not an oversight.
- Closes the "you still need nginx in front of it" gap in the single-binary pitch — extends the honesty work from v0.90.0's static-TLS CA bundle to the server side.
- See issue #260 — STARTTLS wiring into `machin-mail` (a separate repo) is a follow-up dogfood step, not part of this change, so the issue stays open for that.

## Test plan
- [x] `go build ./...` clean
- [x] `go test .` full suite passes
- [x] `TestServerTLS` — a real Go TLS client drives a full handshake + request/response round trip against the raw `tls_accept` builtins
- [x] `TestServeTLS` — the actual `serve_tls` framework function, with a router, driven by a real Go `http.Client` over TLS
- [x] `TestSTARTTLS` — `tls_client_fd` correctly **rejects** an untrusted/self-signed cert (proves verification is active on the upgrade path, not silently disabled)
- [x] Manually verified with `curl -k` against a `serve_tls` router app (full round trip)
- [x] `machin guide` / `--text` render the new `tls` category and idiom correctly; `TestGuideBuiltinsRecognized`/`TestGuideIdiomsCompile` pass

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,27 @@
 
 ## Unreleased
 
+- **Server-side TLS + STARTTLS — machweb can terminate HTTPS itself, no reverse
+  proxy needed.** New builtins `tls_server_ctx(certfile, keyfile) -> int`,
+  `tls_accept(ctx, fd) -> int` (server handshake), `tls_client_fd(fd, hostname)
+  -> int` (the STARTTLS primitive — upgrade an already-connected, plaintext-
+  negotiated fd to verified TLS in place), and `tls_read[_bytes]`/
+  `tls_write[_bytes]`/`tls_close` (mirroring the plain fd read/write/close
+  builtins over a tls handle). New `framework/machweb.src` function
+  `serve_tls(port, certfile, keyfile, handler)` — `serve`'s TLS-terminating
+  sibling, reusing the same router/handler/Response machinery; v1 scope is
+  plain request/response only (`res.is_hijack`/`res.is_stream` get a 501 rather
+  than misbehaving over a tls handle — a documented, deliberate limitation, not
+  an oversight). Verified end-to-end: a real Go TLS client driving a full HTTP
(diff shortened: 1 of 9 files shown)


javimosch:
**Review — Head of Org Engineering:** The change correctly implements server-side TLS termination and STARTTLS primitives as described, with syntactically valid C and Go code, proper scoping to TLS-related files only (codegen.go, framework/machweb.src, CHANGELOG.md, SPEC.md), and no secrets, destructive edits, or correctness bugs — the handshake refactor into mfl_tls_handshake_client is sound and the documented 501 limitation for hijack/stream is deliberate and safe.

PR #288, opened 2026-07-02
proof section gains data-race safety — closes #287

### LINKED ISSUE #287: machin guide proof section: add data-race safety — the flagship claim has zero reproducible evidence
## Gap: the flagship differentiator is invisible to the proof pipeline

v0.91.0 shipped **inferred data-race freedom** (`machin check`'s `race` phase,
`--race-safe`) — the actual headline differentiator of the whole project ("Go's
concurrency, Rust's guarantee, neither's tax"). It shipped with prose docs (a
`gotchas` entry, `commands` descriptions) but was **never added to
`proof.benchmarks`** (v0.89.0), unlike the three benchmarks already there (agent
write cost, native speed, cold-start size) — each of which is reproducible, not
asserted.

That section exists specifically so another agent researching "what should I
build this in" reads `machin guide` and relays structured evidence to its human
principal. Right now it relays three lesser claims fully instrumented and omits
the one claim no other language can match. The asymmetry itself is bad: it reads
as "the benchmarks we could prove, we proved; this one we couldn't."

## What I verified (spiked, not speculated) — and an honest surprise

Canonical shared-counter race: 4 goroutines/threads each incrementing a global
counter 2,000,000 times with no synchronization (expected sum: 8,000,000).

- **machin**: `machin check --json` on the untouched, zero-annotation program
  returns `RACE001` (write/write) with a precise counterexample. `machin build
  --race-safe` refuses (exit 1). Plain `machin build` still succeeds (backward
  compatible) — and on this hardware, running it did **not** visibly corrupt the
  output (printed the correct `8000000`).
- **Go**: `go build` compiles the equivalent silently — no error, no warning.
  Running it **did** visibly corrupt the output (~3,000,000–3,700,000 across
  three runs, never 8,000,000) — a real, observable bug an agent could ship
  without ever knowing. `go run -race` (an opt-in, dynamic instrumentation tool,
  not the compiler) independently confirms it's a genuine data race — but only
  if you remember to reach for it, and only for interleavings the specific run
  happens to hit.
- **Rust**: the naive translation (a shared `static mut`, no wrapper types)
  **fails to compile** — `error[E0133]: use of mutable static is unsafe`, name-
  checking the exact danger. Wrapping the accesses in `unsafe` makes it compile
  (with a further compiler warning) — and, like machin, happened to print the
  correct number on this run. The actually-safe idiomatic fix needs
  `Arc<AtomicI64>` + `Ordering::SeqCst` + `.fetch_add`/`.load` — real additional
  types/syntax over the naive shape, the "tax" the pitch refers to.

**The honest, more interesting story than "look, it crashes":** whether the
racy program's output looks wrong on a given run is unreliable evidence either
way (machin and even `unsafe`-Rust happened to print the correct number this
time; only Go's differently-scheduled runtime visibly corrupted here) — which
is exactly *why* compile-time detection matters. `--race-safe` catches the
defect deterministically, on the untouched original code, independent of
whether that specific run would have manifested visible corruption. That's a
sharper, more defensible claim than a flaky "watch it break" demo, and it's
cross-validated by an independent tool (Go's own `-race` detector agrees this
is a real race).

Confirmed the fix (share-by-communicating: each goroutine returns its partial
sum over a channel instead of touching shared state) passes `machin check`
clean and `--race-safe` builds successfully, zero annotations added — same
code shape Go/machin idiomatically already encourage, just enforced.

## The bounded v1

- `bench/race-freedom/`: `racy.src` (the canonical race) + `safe.src` (the fix)
  in machin, `racy.go` in Go, `racy_no_unsafe.rs` (fails to compile — captures
  the E0133 error text) + `racy_naive.rs` (the `unsafe` escape hatch) +
  `safe.rs` (`Arc<AtomicI64>`) in Rust, a `run.sh` that runs the full sequence
  above and prints a verdict, and a `README.md` telling the honest story
  (including the "numeric output isn't reliable evidence" finding — don't
  oversell it as "look it crashes").
- A 5th `proof.benchmarks` entry in `guide.go`, `axis: "data-race safety"`,
  pointing at `bench/race-freedom` — same `TestGuideProofReproducible` coverage
  every other entry already gets for free.
- New `TestGuideCatalog`/idiom updates only if needed; no compiler changes —
  this is purely closing the proof-section gap for an already-shipped feature.

## Explicitly deferred (not v1)

- A quantitative "annotation count" metric (e.g. lines-of-code diff) — the
  qualitative story (compile error vs silent vs zero-annotation-catch) is the
  point; a token-count-style number can be a follow-up if it turns out to add
  real signal, not just decoration.
- Extending the demo to slice/map races or the closure-capture case — the
  global-counter shape is the most universally recognizable "textbook race"
  and is enough to carry the claim; more shapes are optional depth, not v1.

Labels: enhancement, docs, proof-section

javimosch:
## Summary
- v0.91.0 shipped inferred data-race freedom (`machin check`'s `race` phase, `--race-safe`) — the project's flagship differentiator — but it only got prose docs, never a reproducible benchmark, unlike everything else in `machin guide`'s `proof` section. This closes that asymmetry.
- New `bench/race-freedom`: the same textbook shared-counter race (4 threads/goroutines, 2M increments each, no sync, expected sum 8,000,000) implemented in **machin, Go, and Rust**, with a `run.sh` that runs all 9 steps and asserts each matched its expected outcome.
- **The interesting finding** (empirical, not assumed going in): numeric output alone is NOT reliable evidence of a race either way. machin's racy build printed the *correct* sum on this run, and so did Rust's `unsafe`-escaped version — only Go's differently-scheduled goroutines visibly corrupted here. All three have the identical race (cross-validated via `go run -race` and Rust's own compiler refusing the naive form). That's the actual argument for compile-time detection: a race that happens to look fine in dev/CI/staging is not a race you've ruled out.
- 5th `proof.benchmarks` entry in `guide.go`; `TestGuideProofReproducible` covers the new bench dir automatically (same discipline as every other entry).

## Test plan
- [x] `go build ./...` clean
- [x] `go test .` full suite passes
- [x] `bench/race-freedom/run.sh` run end-to-end — all 9 steps matched expected outcomes (`machin check` flags RACE001, `--race-safe` refuses, Go compiles silently + visibly corrupts, `go run -race` confirms, Rust naive fails E0133, Rust `unsafe` compiles+warns, Rust safe fix works, machin's fix passes `--race-safe` clean)
- [x] `machin guide` / `--text` render the new benchmark row correctly
- [x] `TestGuideCatalog` / `TestGuideProofReproducible` pass

## Summary by CodeRabbit

* **New Features**
  * Added a new benchmark and guide entry for data-race safety comparisons across supported toolchains.
  * Introduced runnable race-freedom examples plus a reproducible verification script.

* **Documentation**
  * Expanded benchmark documentation with scenario details, expected outcomes, and reproduction steps.

* **Bug Fixes**
  * Added a thread-safe version of the benchmark alongside the race-prone examples for clearer comparison.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,24 @@
 
 ## Unreleased
 
+- **`machin guide`'s `proof` section gains a 5th entry: data-race safety —
+  closing the gap where the project's flagship differentiator (v0.91.0's
+  inferred data-race freedom) had prose docs but no reproducible benchmark,
+  unlike everything else in that section.** New `bench/race-freedom`: the same
+  textbook shared-counter race (4 threads, no sync, expected sum 8,000,000) in
+  machin, Go, and Rust. `machin check` catches it at compile time on the
+  untouched code (`RACE001`) and `--race-safe` refuses the build; Go's compiler
+  accepts it silently (and visibly corrupted the output on this run — 3
+  wrong numbers, confirmed a genuine bug via `go run -race`, not a false
+  claim); Rust's naive translation fails to compile (`E0133`), and its actual
+  safe fix needs `Arc<AtomicI64>`/`Ordering` wrapper types machin's
+  zero-annotation channel-based fix doesn't need. The more useful finding than
+  "it crashes": numeric output alone is NOT reliable evidence of a race either
+  way — machin's racy build printed the correct sum on this run too, and so
+  did Rust's `unsafe`-escaped version — which is exactly why compile-time
+  detection matters more than hoping a test run happens to expose the bug.
+  See issue #287.
+
 ## v0.92.0
 
 - **Server-side TLS + STARTTLS — machweb can terminate HTTPS itself, no reverse
--- a/bench/race-freedom/.gitignore
+++ b/bench/race-freedom/.gitignore
@@ -1,0 +1,10 @@
+# built artifacts — the source + run.sh are the benchmark
+racy.mfl
+safe.mfl
+racy-race-safe
+racy-machin
+racy-go
+racy_no_unsafe
+racy_naive
+safe-rust
+safe-machin
--- a/bench/race-freedom/README.md
+++ b/bench/race-freedom/README.md
@@ -1,0 +1,55 @@
+# Benchmark — data-race safety, three ways
+
+machin v0.91.0 shipped **inferred data-race freedom**: `machin check` infers,
+with zero annotations, whether a program's goroutines can race — the guarantee
+Rust needs `Send`/`Sync` for. This benchmark puts that claim next to Go and
+Rust on the same textbook example. See issue #287.
+
+The program: 4 threads/goroutines each increment a shared counter 2,000,000
+times with no synchronization. Expected sum: 8,000,000. A data race on the
+counter means the true answer is undefined behavior.
+
+## Results (this machine — gcc 11, go 1.22, rustc 1.98)
+
+| | compiles? | catches the race? | how | numeric output (3 runs) |
+|---|---|---|---|---|
+| **machin** (`racy.src`) | yes | **yes — `machin check`, at compile time** | `RACE001` on the untouched code; `--race-safe` refuses the build | 8000000 / 8000000 / 8000000 |
+| Go (`racy.go`) | yes, silently | only via `go run -race` (opt-in, dynamic) | no error, no warning from `go build` | 2094040 / 3638549 / 2090454 |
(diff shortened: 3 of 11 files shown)


javimosch:
**Review — Head of Org Engineering:** The PR adds a new bench/race-freedom directory with benchmark sources, a run script, README, and a CHANGELOG entry — exactly what the title/description describes. All files are syntactically valid, scoped to the new benchmark directory, contain no secrets or destructive changes, and touch no declared no-touch areas.

Issue #280, opened 2026-07-01
Self-hosted codegen: concurrency (go/chan/select) + closures — reach 100% drop-in parity


javimosch:
## Goal
Bring the **self-hosted** compiler (`selfhost/`) to 100% parity with the Go reference by
supporting the two feature families it currently can't codegen: **concurrency**
(`go` / channels / `select`) and **closures** (`func` literals / func-valued variables).
The self-hosted `machin` is already a byte-exact drop-in for everything else (libc, crypto,
SQLite, math, TLS, regex, FFI/3D, and itself). These are the last gaps.

This is a **completionist milestone**, not a user-facing blocker: every affected program
already builds fine with the reference compiler. Tracking it so it can be done deliberately.

## What's blocked (corpus scan)
`go` appears in 8 of the 9 unsupported apps; `func(` closures in 4 (the machweb web apps).
- **concurrency**: machin-healthcheck, machin-linkcheck, machin-pipe, machin-pool, machin-wscat
- **closures**: machin-meet, machin-notify, machin-web-demo-ssr, machin-web-demo-reactive

## Method (unchanged)
Same oracle-diff discipline as the rest of the bootstrap: each construct's emitted C is
byte-diffed against `machin cgentest --program` (and `--emit-c` for the full path). A slice
is done only when the diff is empty across the corpus. The checker already accepts `go`
(`selfhost/checkgen.src`); the gaps are in codegen (`selfhost/cgen.src:204,372`,
`selfhost/cgagg.src:227`) and the checker's `select`/funclit paths (`checkgen.src:920`).

## Concurrency (sliced)
The `go` codegen is the structural crux: it emits a per-call arg **struct + trampoline
function into a separate top-level buffer** (like the ret-structs), then a detached
`pthread_create` at the call site. The channel runtime (`mfl_make_chan`/`mfl_chan_send`/
`mfl_chan_recv2`) is already in the core prelude — only the codegen is missing.

- [ ] **Slice 1 — `go` + scalar channels.** Trampoline top-level buffer + `go f(args)` call
      site; `make(chan int/bool)` → `mfl_make_chan(sizeof(T),0,0,0)`; `ch <- v` →
      `mfl_chan_send`; `<-ch` (single + comma-ok, the latter already ported) via a hoisted
      temp. Verify emitted C vs Go on a hand-written scalar-channel program.
- [ ] **Slice 2 — string channels + `range` over a channel.** String-offset marshaling in
[...]

Issue #208, opened 2026-06-24
lexer: `a < -b` mis-tokenizes as channel-receive (`<-`) after encode tightening


javimosch:
## Repro

```
func main() { x := 5  if x < -1 { println("neg") } else { println("ok") } }
```

```
$ machin encode trap.src
error: parse: expected "{", got "<-" at pos 21
```

## Cause

`encode` tightens whitespace adjacent to operators, so `x < -1` becomes `x<-1`. The lexer then greedily matches `<-` as the **channel-receive** token, and the parser sees `if x <- 1 { ... }`. So a perfectly valid comparison against a negative value fails to parse **after the round-trip through canonical form** — i.e. tightening changes meaning.

It bites any `a < -b`: negative thresholds (`if h < -0.7`), `i < -1`, etc. Hit while writing [machin-demo-terrain](https://github.com/javimosch/machin-demo-terrain) (`if ya < -0.7`).

## Workaround

Write `a < 0 - b` (or `a < (0 - b)`), or flip the comparison so `<` isn't followed by `-`. `if x < 0 - 1 { ... }` encodes and runs fine.

## Suggested fix

`<-` should only be the receive token where a receive is grammatical (a unary/prefix or statement position), not between two expressions. Options:
- **Lexer:** don't emit `<-` when the previous significant token ends an expression (an identifier, literal, `)`, or `]`); emit `<` then `-` instead. (Mirrors how `>>`-style ambiguities are sometimes split.)
- **Parser:** in infix position, treat a `<-` token as `<` followed by unary `-` (re-split).

Either keeps `v := <-ch` and `ch <- v` working while letting `a < -b` parse. A regression test on the encode→parse round-trip of `a < -b` would lock it in.

Low-severity (clean workaround exists) but it's a correctness wart: the canonical form of a valid program fails to parse.

javimosch:
Fixed in the reference (Go) compiler — see the PR. Root cause was slightly different from either suggested fix: since \`ch <- v\` (a genuine channel send) is lexically the identical shape to \`x < -1\` after tightening (an identifier immediately followed by \`<-\`), a pure lexer-level fix can't disambiguate them — it depends on grammatical position (statement-start vs. mid-expression), which only the parser knows. Fixed there: the one call site that recognizes send statements still stops at a bare \`<-\`; everywhere else now reinterprets it as \`<\` + unary \`-\`.

[...]

PR #289, opened 2026-07-02
fix: a < -b failed to parse after its own canonical form (#208)

javimosch:
## Summary
- `machin encode` tightens whitespace around operators, so `x < -1` becomes byte-adjacent `x<-1` in canonical form — and the lexer greedily merged that into a single channel-receive `<-` token, so a valid comparison against a negative literal failed to parse *after* the round trip through canonical form (the generate → canonicalize → re-ingest loop an agent runs constantly).
- Root cause is slightly different from either fix the issue sketched: `ch <- v` (a genuine channel send) is lexically the **identical** shape to `x < -1` after tightening (an identifier immediately followed by `<-`), so a lexer-only fix can't disambiguate them — it depends on grammatical position (statement-start vs. mid-expression), which only the parser knows.
- Fixed in the parser: `parseExprForStmt` (used only for a simple statement's leading expression, which may turn out to be a channel send) still stops at a bare `<-`, exactly as before. Every other caller (`if`/`while` conditions, assignment/return values, call arguments, nested sub-expressions) now goes through `parseExpr` → `parseBinaryFrom`, which reinterprets an unexpected `<-` found while precedence-climbing as `<` followed by unary `-` on the next operand.
- **Not fixed here** (flagged as a follow-up on #208): `selfhost/parse.src` has its own independent precedence-climbing implementation with the same `<-`-handling shape, so it likely has the identical bug — left alone to avoid colliding with the in-flight self-hosted concurrency-parity work (#280).

## Test plan
- [x] `go build ./...` clean
- [x] `go test .` full suite passes (`-count=1`, no cache)
- [x] New `TestLessThanNegative`: the reported case, plus operator-precedence preservation (`a && b < -1` parses as `a && (b < -1)`, not `(a && b) < -1`) and composition with further arithmetic (`c < -1 + 10`)
- [x] New `TestLessThanNegativeChannelSendUnaffected`: the exact same ambiguous token shape (`ch <- v`) still correctly recognized as a send, not split
- [x] Manually verified round-trip idempotency (re-encoding the canonical form is a no-op) and `select { case <-done: }` still works

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,33 @@
 
 ## Unreleased
 
+- **Fixed: `a < -b` failed to parse after its own canonical form (issue #208).**
+  `machin encode` tightens whitespace around operators, so `x < -1` becomes
+  byte-adjacent `x<-1` — and the lexer greedily merged that into a single
+  channel-receive `<-` token, so a valid comparison against a negative literal
+  failed to parse *after* the round trip through canonical form (the exact
+  loop-integrity contract — generate, canonicalize, re-ingest — an agent runs
+  constantly). Since `ch <- v` (a genuine channel send) is lexically the
+  identical shape (an identifier immediately followed by `<-`), this couldn't
+  be fixed at the lexer: whether `IDENT <- ...` means "send" or "less-than,
+  then negate" depends on grammatical position, not local token context. Fixed
+  in the parser instead: the one call site that recognizes send statements
+  (a simple statement's leading expression) still stops at a bare `<-` as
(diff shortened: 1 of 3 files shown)


javimosch:
**Review — Head of Org Engineering:** The change correctly fixes the `<-` greedy-merge bug by splitting `parseBinary` into `parseExprForStmt`/`parseBinaryFrom` and reinterpreting unexpected `<-` as `<` plus unary `-` in non-send contexts. The diff is syntactically valid Go, scoped to `parser.go`, `mfl_test.go`, and `CHANGELOG.md`, with no secrets, destructive edits, or out-of-scope changes.

PR #290, opened 2026-07-02
fix: machin run/build rejected hand-written loose MFL source

javimosch:
## Summary
- Dogfooded 5 parallel demos targeting the least battle-tested recent features: `serve_tls` in a real REST+SQLite API, `--static` + `serve_tls` + SQLite combined (never tried together before — verified even in a genuine `FROM scratch` Docker container), `tls_client_fd` doing a real STARTTLS negotiation against `smtp.gmail.com`, `--race-safe` against a deliberately-introduced shared-map bug in a concurrent file-hasher, and secp256k1/keccak256 used generically (keygen/sign/verify) outside the EIP-712 lens.
- **Every one of those features came back clean** under real, adversarial testing — but **3 of 5 independent agents hit the identical friction**: writing a normal multi-line, Go-like `.mfl` file and running it directly failed with `"line is neither plain MFL nor base64: illegal base64 data at input byte N"`. `loadMFL` only understood canonical one-declaration-per-line text or the packed base64 form — a line like `println(x)` has no whitespace, so it looked like a malformed packed declaration instead of a fragment of a multi-line function body.
- `machin check` already tolerated this shape (`analyzeSource` uses the same encode-style `splitFunctionsLoc`/`normalize` machinery `machin encode` does) — `loadMFL` now does too, via a new `loadDecls` helper, closing the **inconsistency between `check` and `run`/`build`** rather than just improving the error message. Packed-base64 detection is unchanged (a file is only treated as packed if EVERY non-blank line lacks whitespace).
- Two smaller fixes found by the same exercise: `serve`/`serve_tls`'s startup banner never called `flush()` (silently lost under any redirected/piped/daemonized deployment); `machin guide`'s `no-tls-without-https` gotcha was stale, contradicted by `tls_client_fd`/`tls_server_ctx`/`tls_accept` shipped two releases ago.

## Test plan
- [x] `go build ./...` clean
- [x] New `TestLoadMFLAcceptsLooseSource` / `TestLoadMFLStillAcceptsCanonicalAndPacked`
- [x] Re-verified the exact scenario that tripped 3 dogfood agents: `machin run`/`build` now accept hand-written multi-line source directly
- [x] Batch-ran the whole `examples/` corpus (39 files) — zero load/parse regressions
- [x] Verified a composed `framework/machweb.src` + app build still works
- [x] Confirmed the packed-base64 round trip (`machin pack` → `machin run`) still works
- [x] Confirmed the `flush()` fix: startup banner now appears with stdout redirected to a file (previously silently lost)


javimosch:
**Review — Head of Org Engineering:** The PR does exactly what it describes: fixes loadMFL to accept loose multi-line source, adds flush() after serve/serve_tls banners, and corrects the stale guide gotcha. All edits are syntactically valid, scoped to the affected files (main.go, cli_test.go, framework/machweb.src, guide.go, CHANGELOG.md), and contain no correctness bugs, secrets, or destructive changes.

PR #291, opened 2026-07-02
feat: self-hosted concurrency — go/chan/select codegen + the data-race pass, in MFL (#280)

javimosch:
Brings the **self-hosted** MFL compiler to concurrency parity (#280) and ports the **data-race analysis** into it — so machin-in-machin both *compiles* concurrent programs and *proves* them race-free.

## Concurrency codegen (#280)
The reference already supported `go`/channels/`select`; this teaches the self-hosted compiler (`selfhost/*.src`) to emit byte-identical C for it:
- **go + scalar channels** — `make(chan T)`, `ch <- v`, `<-ch`, and the `go f(args)` trampoline (arg-struct + trampoline fn spliced before bodies via a body-buffer swap) + detached `pthread_create`.
- **string channels + `range` over a channel** (`mfl_chan_recv2` loop).
- **`select`** — checker `K_SELECT` constraints + the poll-loop codegen.
- **struct channels** — the `offsetof` string-copy path (`chan_needs_json`/`chan_str_offsets`).

**Result: all 5 concurrency corpus apps (healthcheck/linkcheck/pipe/pool/wscat) now self-host byte-identically.** `verify-cgen.sh` 411 PASS / 0 FAIL (incl. fuzz + self-application).

## The data-race pass, in MFL
Ported `racecheck.go` (the "better than Rust" inferred race checker) into `selfhost/racecheck.src`, verified against a new `machin racetest --program` oracle. Flat-rewritten (MFL has no closures): local param races (type-aware reachability), happens-before (live-counting + channel-join barriers + loop multiplicity), package globals (unconditional sharing), and move-on-send (use-after-move). `verify-race.sh` 22 PASS / 0 FAIL — byte-identical to the reference, and the race-free corpus stays clean.

Only slice/map (JSON) channels remain of #280 codegen — no corpus program uses them.

`go test .` green.

## Summary by CodeRabbit

* **New Features**
  * Added a new `racetest` command for generating canonical race-analysis output from a program file.
  * Expanded self-hosted support for channels, `select`, `go` routines, and channel ranges.

* **Bug Fixes**
  * Improved race detection coverage, including shared-state races and use-after-send/move cases.
  * Added clearer parse/check error reporting for race test runs.

* **Tests**
  * Added broader verification coverage for concurrency, channels, and race-analysis behavior.

### DIFF
--- a/main.go
+++ b/main.go
@@ -48,6 +48,12 @@ func main() {
 			os.Exit(1)
 		}
 		return
+	case "racetest":
+		if err := cmdRaceTest(os.Args[2:]); err != nil { // self-hosting oracle: dump race findings canonically
+			fmt.Fprintln(os.Stderr, "error:", err)
+			os.Exit(1)
+		}
+		return
 	case "uftest":
 		if err := cmdUFTest(os.Args[2:]); err != nil {
 			fmt.Fprintln(os.Stderr, "error:", err)
--- a/racetest.go
+++ b/racetest.go
@@ -1,0 +1,53 @@
+package main
+
+import (
+	"encoding/hex"
+	"fmt"
+	"os"
+	"strings"
+)
+
+// cmdRaceTest is the self-hosting oracle for the data-race pass:
+//
+//	machin racetest --program <file.mfl>
+//
+// It runs parse → liftClosures → check → detectRaces and dumps every finding in a
+// canonical, escaping-free form (one per line, fields hex-encoded, in detectRaces'
+// stable sort order) so the MFL port can be byte-diffed against it.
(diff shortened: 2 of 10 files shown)


javimosch:
**Review — Head of Org Engineering:** The diff adds a new `racetest` subcommand (main.go + racetest.go) and implements channel codegen (make/send/recv/select/go) plus supporting helpers in selfhost/cgen.src and selfhost/cgagg.src. All changes are syntactically valid, scoped to the channel-support feature, and contain no secrets, destructive edits, or correctness bugs visible in the diff.

PR #294, opened 2026-07-02
fix: read_file/read_file_bytes segfaulted on a directory path

javimosch:
## Summary
- `fopen(dir, "rb")` succeeds on Linux (opening a directory read-only is legal at the syscall level), but `ftell()` on the resulting stream returns `LONG_MAX` — not `-1` — so the existing `if (n < 0) n = 0` guard never caught it, and the runtime tried to allocate ~9.2 exabytes for the "file size."
- A very easy path to hit in practice: `list_dir()`'s entries can themselves be subdirectories, so `read_file_bytes` on any of those crashed. Found while building a concurrent file-hasher demo (`machin-hasher`) that reliably segfaulted on any target directory containing so much as one subdirectory — reproduced deterministically even with a single worker, so it was never a race, just a plain crash waiting for the first subdirectory.
- Fixed with an explicit `stat()`-based directory check before `fopen` in both `mfl_read_file` and `mfl_read_file_bytes`; both now return empty (matching their existing "can't open" behavior) instead of crashing.

## Test plan
- [x] `go build ./...` clean
- [x] New `TestReadFileOnDirectory` — both builtins return empty on a directory path instead of crashing
- [x] Reproduced the original crash manually (a directory with one file + one subdirectory, worker count 1 — deterministic, not a race) and confirmed the fix resolves it

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,18 @@
 
 ## Unreleased
 
+- **Fixed: `read_file`/`read_file_bytes` segfaulted on a directory path.** `fopen(dir,
+  "rb")` succeeds on Linux (opening a directory read-only is legal at the syscall
+  level), but `ftell()` on the resulting stream returns `LONG_MAX` — not `-1` — so the
+  existing "if (n < 0) n = 0" guard never caught it, and the runtime tried to allocate
+  ~9.2 exabytes for the "file size." A very easy path to hit in practice: `list_dir()`'s
+  entries can themselves be subdirectories, so `read_file_bytes` on any of those crashed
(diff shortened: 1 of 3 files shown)


javimosch:
**Review — Head of Org Engineering:** The fix correctly addresses the described segfault by adding a `stat()`-based directory check before `fopen` in both `mfl_read_file` and `mfl_read_file_bytes`, matching the PR description exactly. The code is syntactically valid C, scoped only to relevant files (codegen.go, cli_test.go, CHANGELOG.md), and introduces no security issues, destructive changes, or secrets.

PR #303, opened 2026-07-03
fix: reject duplicate function definitions (last-wins was silent)

### LINKED ISSUE #88: bug: duplicate function/type definitions are silently accepted (last wins) — silent shadowing in encode composition
## Problem

Defining two functions (or two types) with the same name does not produce any error. The compiler silently keeps the **last** definition and discards the earlier one — no warning, no diagnostic.

## Reproduction

```
$ printf 'func greet() { return "FIRST" }\n\nfunc greet() { return "SECOND" }\n\nfunc main() { println(greet()) }\n' > dup.src
$ machin encode dup.src > dup.mfl   # exit 0, no warning
$ machin run dup.mfl
SECOND
```

`encode` runs `ParseProgram` + `Check` (see `cmdEncode` in `main.go`) and both pass; the duplicate is never flagged.

## Why it matters

This is a real footgun for the framework composition model, which is the documented way to build apps:

```
machin encode framework/machweb.src app.src > app.mfl
```

`encode` concatenates the framework and the app into **one global namespace**. If an app function/type name collides with one in `machweb.src` (`response`, `route`, `param`, `serve`, `dispatch`, `Request`, `Response`, …), one definition silently overrides the other and the program compiles to something the author did not write.

This already exists latently in the repo's own shipped example: `framework/example.src` defines `func route(req)`, while `framework/machweb.src` defines `func route(router, method, path, handler)`. Composing them yields two `route` definitions; it only works today because `example.src` calls `serve()` (not `serve_router()`), so the framework's `route` is never invoked. A small refactor would turn this into a baffling bug.

## Suggested fix

- In the type checker (`Check` in `types.go`), reject a duplicate top-level function or type name with a clear `error: function "route" redefined` diagnostic.
- Until then, rename `example.src`'s `route(req)` (e.g. to `handle_request`) and document in `framework/README.md` that composed `.src` files share a single global namespace, so app names must not collide with machweb's.

javimosch:
## Summary
- Two functions with the same name compiled with no error — `Check`'s registration loop just overwrote the earlier entry in its lookup map. Types already had this guard (`duplicate type %q`); functions had none.
- A real footgun for the framework composition model, which puts the framework and the app in one shared global namespace: an app function name colliding with a framework one silently won or lost depending on declaration order.
- Added the matching guard, classified as `duplicate-function` in `machin check --json`'s stable code list.
- Fixing this surfaced a real latent collision already shipping in the repo: `framework/example.src`'s `route(req)` collided with `machweb.src`'s `route(router, method, path, handler)` — `framework/run.sh` composes both by default. Renamed to `handle_request`; verified the example still serves correctly end-to-end. `framework/README.md` now documents the shared-namespace rule explicitly.

Fixes #88.

## Test plan
- [x] `go build ./...` clean
- [x] New `TestCheckDuplicateFunction`
- [x] Reproduced the original bug (`greet()` defined twice → `SECOND` silently wins), confirmed it now errors with `duplicate function "greet"` / `machin check --json` code `duplicate-function`
- [x] `framework/run.sh` (composes `machweb.src` + `example.src`) verified working end-to-end after the rename — curl against `/`, `/api/todos`, `/hello/machine` all correct
- [x] Checked all other `machin encode` compose call sites in the repo (`bench/*`) for the same collision class — none found

## Summary by CodeRabbit

* **Bug Fixes**
  * Duplicate function names are now rejected with a clear diagnostic instead of being accepted silently.
  * Analysis now reports a dedicated error code for this case, improving JSON output and tooling feedback.

* **Documentation**
  * Updated docs to explain that app and framework code share one namespace, and that name collisions are not allowed.

* **Tests**
  * Added coverage for duplicate function detection to verify the new error behavior.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,21 @@
 
 ## Unreleased
 
+- **Fixed: duplicate function definitions were silently accepted (last wins).** Two
+  functions with the same name compiled with no error — the checker's registration
+  loop just overwrote the earlier entry in its lookup map. A real footgun for the
+  framework composition model (`machin encode framework/machweb.src app.src`), which
+  puts the framework and the app in one shared global namespace: an app name
+  colliding with a framework name (`route`, `param`, `serve`, …) silently won or lost
+  depending on declaration order, with zero diagnostic. `Check` already rejected
+  duplicate *types* this way; functions had no equivalent guard. Now a clear
+  `duplicate function "name"` error (`machin check --json` code `duplicate-function`).
+  Fixing this surfaced a real latent collision already shipping in the repo:
+  `framework/example.src`'s `route(req)` collided with `machweb.src`'s
+  `route(router, method, path, handler)` — renamed to `handle_request`, and
+  `framework/README.md` now documents the shared-namespace rule explicitly. See
+  issue #88.
+
 - **Added: `machin test [--json] <src...>` — a native MFL test runner (#236 Stage A).**
   `framework/test.src` provides `assert(cond, msg)`, `assert_eq_int(got, want, name)`,
   `assert_eq_str(got, want, name)`, and `test_summary()` (call last in `main()` — prints the
--- a/check.go
+++ b/check.go
@@ -349,6 +349,8 @@ func classifyCheck(msg string) string {
 		return "no-main"
 	case strings.Contains(msg, "duplicate type"):
 		return "duplicate-type"
+	case strings.Contains(msg, "duplicate function"):
+		return "duplicate-function"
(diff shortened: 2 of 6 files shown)


javimosch:
**Review — Head of Org Engineering:** The change correctly implements duplicate function detection in the checker (types.go), wires the new error code into classifyCheck (check.go), adds a focused regression test (check_test.go), renames the colliding example function (framework/example.src), and documents the shared-namespace rule (framework/README.md, CHANGELOG.md). All edits are in-scope, syntactically valid, and free of correctness bugs or security issues.

PR #304, opened 2026-07-03
fix: json_echo_api.mfl silently truncated large/split POST bodies (#91)

### LINKED ISSUE #91: bug: read() does a single 64KB read — large/split HTTP requests are silently truncated in the showcased servers
## Problem

The `read` networking builtin issues a single `read(2)` of at most 65535 bytes and returns whatever that one call produced:

```c
// codegen.go:364-370
static char* mfl_read(int64_t fd) {
    char* buf = mfl_alloc(65536);
    ssize_t n = read((int)fd, buf, 65535);
    if (n < 0) n = 0;
    buf[n] = 0;
    return buf;
}
```

TCP is a byte stream: a single `read()` is **not** guaranteed to return a whole HTTP request. A request larger than ~64KB, or one whose bytes arrive in multiple segments (common for POST bodies, or under load), will be **silently truncated** — `read()` returns a partial buffer with no error.

Every showcased server relies on one `read(conn)` returning the full request:

- `framework/machweb.src:62-67` — `machweb_handle` does `raw := read(conn)` then `parse_request(raw)` / `http_body(raw)`.
- `examples/complex/http_server.mfl`, `examples/complex/json_echo_api.mfl` (POST → parse body → echo), and `selfhost/server.mfl` all assume a complete request from one read.

So the headline "POST JSON → parse into a struct → echo it back" feature (`json_echo_api`, framework echo route) can mis-parse or drop data whenever the body doesn't fit in the first segment.

## Why it matters

The concurrent HTTP server and JSON-over-HTTP APIs are flagship examples of the language. A partial read produces a malformed request line or a truncated JSON body, which then fails `parse(...)` or yields wrong results — intermittently, and only under conditions (size/timing) that don't show up in a quick local `curl`.

## Suggested fix

- Loop on `read()` until the request is complete (e.g. headers terminated by `\r\n\r\n`, then read `Content-Length` bytes of body), growing the buffer as needed; or at minimum loop to drain what's currently available.
- If a full fix is out of scope, document the 64KB / single-read limitation alongside `read` in `docs/LANGUAGE.md` (§Builtins, line ~330) and `SPEC.md` §8 so users of `listen`/`accept`/`read` know requests are not reassembled.

javimosch:
## Summary
- `read(fd)` is one `read(2)` syscall, returning at most 65535 bytes — not a whole HTTP request. `framework/machweb.src`'s own server (`read_request_bytes`) already loops correctly on this; `examples/complex/json_echo_api.mfl` — one of the two examples #91 named as actually vulnerable (`http_server.mfl` reads but never uses the result, so it's not functionally affected) — still used a single `read(conn)`.
- Fixed by inlining the same loop-until-`Content-Length` pattern (`read_bytes` + `bytes_index` + a small `content_length` helper mirroring `machweb.src`'s), kept self-contained so the example needs no framework composition.
- `docs/LANGUAGE.md` (new "Raw sockets" subsection) and `SPEC.md` now document the single-read limitation next to `read`/`write`, pointing at `read_request_bytes` as the reference pattern — per the issue's own documentation fallback.

Fixes #91.

## Test plan
- [x] `go build ./...` clean
- [x] New `TestReadBytesLoopReassemblesLargePayload` — proves the underlying mechanism (a single `read_bytes` call cannot return more than its 65536-byte cap) a regression could silently break
- [x] Empirically reproduced the bug against a real 100 KB POST body: old code truncated to 65,353 bytes; fix round-trips all 100,000 — verified with a clean server process each time (not a stale one), confirmed by explicit port checks
- [x] Canonical-form check (the file's own tightening rules) passes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,18 @@
 
 ## Unreleased
 
+- **Fixed: `examples/complex/json_echo_api.mfl` silently truncated large/split POST
+  bodies.** `read(fd)` is one `read(2)` syscall, returning at most 65535 bytes — not a
+  whole HTTP request. `framework/machweb.src`'s own server (`read_request_bytes`) has
+  looped correctly on this for a while; this specific example still used a single
+  `read(conn)` and had no test proving otherwise. Fixed by inlining the same
+  loop-until-`Content-Length` pattern (`read_bytes` + `bytes_index` + a small
+  `content_length` helper), kept self-contained (no framework composition needed).
+  Verified against a real 100 KB body: the old code silently truncated to 65,353
+  bytes; the fix round-trips all 100,000. `docs/LANGUAGE.md` and `SPEC.md` now
+  document the single-read limitation next to `read`/`write`, pointing at the correct
+  pattern. New `TestReadBytesLoopReassemblesLargePayload`. See issue #91.
+
 ## v0.99.0
 
 - **Fixed: duplicate function definitions were silently accepted (last wins).** Two
--- a/SPEC.md
+++ b/SPEC.md
@@ -325,7 +325,7 @@ startup (before `main`; at `_initialize` for a wasm reactor).
 | `sleep` | `(int) -> ` | pause (milliseconds) |
 | `dial` | `(string, int) -> int` | connect to host:port; an fd, or -1 on failure |
 | `listen`, `accept` | `(int) -> int` | open / accept on a TCP socket |
-| `read`, `write` | `(int[, string]) -> string\|int` | socket/fd I/O |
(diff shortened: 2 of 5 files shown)


javimosch:
**Review — Head of Org Engineering:** The PR correctly fixes the single-read truncation bug in examples/complex/json_echo_api.mfl by inlining a read loop matching the framework pattern, updates SPEC.md and docs/LANGUAGE.md with accurate documentation of the read(2) limitation, and adds a concrete regression test (TestReadBytesLoopReassemblesLargePayload) that verifies multi-read reassembly of a 100 KB payload. No correctness bugs, security issues, out-of-scope edits, or leaked secrets are present.

PR #307, opened 2026-07-03
ci: add a workflow that actually runs the test suite (#85)

### LINKED ISSUE #85: ci: no workflow runs the test suite — releases are published without running `go test`
## Problem

The only GitHub Actions workflow is `.github/workflows/release.yml`, which triggers on `v*` tags and **only builds/cross-compiles binaries** — it never runs the test suite. There is no `push`/`pull_request` workflow at all.

Consequences:
- The ~69 Go tests in `mfl_test.go` and `operators_net_test.go` (exercising the full language surface via the native compile path) **never run in CI**. Regressions can land on `main` undetected.
- `release.yml` publishes binaries to a GitHub Release on every `v*` tag **without first running `go test`** — a broken compiler can be shipped to users.

## Where

- `.github/workflows/release.yml` — the build job (`jobs.release`) goes straight from `setup-go` to `go build` with no test step.
- `Makefile` already has a `test` target (`go test ./...`), so the command exists; nothing invokes it in CI.

## Why it matters

This is a compiler: a silent miscompilation or type-inference regression is high-impact and exactly what the existing test suite is designed to catch. Running it automatically on PRs/pushes (and gating releases on it) is cheap insurance.

## Suggested fix

1. Add a `ci.yml` workflow triggered on `push` and `pull_request` that runs `go build ./...`, `go test ./...`, and ideally `./examples/run.sh` (it builds `machin` and runs every non-server example, so it doubles as an integration smoke test). Note it needs a C compiler on the runner — `ubuntu-latest` ships `cc`.
2. Add a `go test ./...` step to `release.yml` **before** the "Build binaries" step so a tag with failing tests does not publish a release.

javimosch:
## Summary
- `release.yml` (build + cross-compile + publish) never ran `go test` — a broken compiler could ship to users, and the ~90 Go tests exercising the full language surface never ran in CI at all.
- Added `ci.yml`: push to `main` + every PR runs `go build`, `go test ./...`, and `./examples/run.sh` (an integration smoke test — builds `machin` and runs every non-server/non-GUI example).
- `release.yml` now runs `go test ./...` before cross-compiling, so a tag with failing tests does not publish a release.
- Both workflows install `libssl-dev`/`libsqlite3-dev` first — the crypto and sqlite builtins' tests link against these, and `release.yml`'s existing cross-compile step (`CGO_ENABLED=0`, pure Go) never needed them before.

Fixes #85.

## Test plan
- [x] `go build ./...` and `go test ./...` pass locally
- [x] Opening this PR triggers `ci.yml` for real — watching the actual GitHub Actions run to confirm the runner environment (system deps, `examples/run.sh`) behaves as expected, not just assuming it will

## Summary by CodeRabbit

* **Chores**
  * Added automated CI checks for pushes and pull requests, including build, test, and a smoke test run.
  * Updated the release workflow to install required system libraries and run tests before publishing binaries.

* **Documentation**
  * Updated the changelog to reflect the new CI and release validation steps.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,33 @@
+name: ci
+
+# Runs on every push to main and every PR: build, the full Go test suite
+# (compiles + executes real MFL programs natively — needs a C compiler and
+# CGO enabled, unlike release.yml's cross-compiled CGO_ENABLED=0 binaries),
+# and every non-server/non-GUI example as an integration smoke test.
+
+on:
+  push:
+    branches: [main]
+  pull_request:
+
+jobs:
+  test:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+
+      - uses: actions/setup-go@v5
+        with:
+          go-version-file: go.mod
+
+      - name: Install system deps (crypto + sqlite builtins link against these)
+        run: sudo apt-get update && sudo apt-get install -y libssl-dev libsqlite3-dev
+
+      - name: Build
+        run: go build ./...
+
+      - name: Test
+        run: go test ./...
+
+      - name: Run every example (integration smoke test)
+        run: ./examples/run.sh
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -23,6 +23,12 @@ jobs:
         with:
           go-version-file: go.mod
 
+      - name: Install system deps (crypto + sqlite builtins link against these)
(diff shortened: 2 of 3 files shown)


javimosch:
**Review — Head of Org Engineering:** The diff does exactly what the PR title/description says: adds a ci.yml workflow that runs build, go test, and example smoke tests on push/PR, and adds go test to release.yml before cross-compiling. All files touched are in-scope (.github/workflows/, CHANGELOG.md), the YAML is structurally valid, and there are no correctness bugs, secrets, or destructive changes.

PR #321, opened 2026-07-04
fix(#156): document json_get in LANGUAGE.md and add a runnable example

javimosch:
Automated maintenance run by automaintainer.

**Focus:** ISSUE FIX OBJECTIVE (GitHub Issue #156: "docs: json_get builtin (jq-style JSON path) is undocumented in docs/LANGUAGE.md and has no runnable example")

ISSUE DESCRIPTION:
## Problem

`json_get(json, path)` is a real, shipped builtin (added in the v0.13.0 era per `CHANGELOG.md`). It is documented in:

- `README.md` line 48 — listed under **Capabilities → I/O & data**: `json`/`parse`/`json_get` (jq-style path)
- `SPEC.md` line 248 — builtins table: `json_get | (string, string) -> (string, string) | value at a jq-style path → (value, err)…`
- the `machin guide` catalog (`guide.go`)

But it is **missing from `docs/LANGUAGE.md`** entirely:

- The **Builtins** table (LANGUAGE.md lines ~311–339) lists `json(x)` and `parse(s, T{})` but not `json_get`.
- The **JSON** subsection (lines ~353–387) walks through `json()` and `parse()` in detail but never mentions `json_get`.

There is also **no runnable example** demonstrating it. `json()` and `parse()` each have dedicated examples (`examples/complex/json.mfl`, `examples/complex/json_parse.mfl`), but a grep of `examples/` shows no `.mfl`/`.src` file uses `json_get` — so an agent reading LANGUAGE.md/examples has no way to learn it exists, despite the README promoting it.

> Note: this is **not** covered by #148 (which is about omitted builtin *categories*: file I/O, HTTP/TLS, time, process, WebSocket) — `json_get` is a JSON builtin sitting next to the `json`/`parse` that *are* documented.

## Why it matters

`json_get` is the idiomatic way to dig into JSON without fragile substring search (the CHANGELOG says "every machin tool used to dig into JSON with fragile substring search" before it). Leaving the canonical reference (and every example) silent on it pushes agents back to the brittle pattern it was added to replace.

## Where

- `docs/LANGUAGE.md` — Builtins table (~line 311) and the JSON subsection (~line 353).
- `examples/complex/` — no example exercises `json_get`.

## Suggested fix

1. Add `json_get` to the LANGUAGE.md Builtins table and document it in the JSON subsection, including the multi-assign `(value, err)` contract and the `err` values (`""`/`"notfound"`/`"path"`/`"parse"`) already specified in SPEC.md line 248.
[...]

Issue #150, opened 2026-06-22
docs: LANGUAGE.md "Control flow" omits break/continue (and the bare `for {}` loop) — they are fully implemented


javimosch:
## Problem
`docs/LANGUAGE.md` is described in the README as the "Language tour" / language reference, but its **Control flow** section (lines 179–202) documents only `if`/`else`, `while`, `for cond`, and `for ... range`. It never mentions `break` or `continue`, nor the bare infinite-loop form `for { ... }`.

These are not unsupported features — they are fully implemented end to end:

- `lexer.go:33` — `break`/`continue` are reserved keywords
- `parser.go:433-437` — parsed into `BreakStmt` / `ContinueStmt`
- `ast.go:185-187`, `types.go:1002-1004`, `codegen.go:1425-1427` — carried through every pass
- `guide.go` lists them as keywords and uses `for { ... if ... { break } }` in its catalog idioms
- `SPEC.md` documents them (§3 keyword list line 82; §7 line 210 "`break` exits the innermost loop; `continue` skips to its next iteration"; grammar line 413)
- `README.md:45` advertises them under **Flow**

So `LANGUAGE.md` is the one reference surface that omits a documented, shipped feature.

## Why it matters
LANGUAGE.md is linked from the README as the language tour; an agent or human reading only it would conclude MFL has no early loop exit. This also makes the doc inconsistent with SPEC.md and the README.

(Note: open issue #90 asserts break/continue "don't exist" — that premise is now stale; this issue is about the LANGUAGE.md omission given that they *do* exist.)

## Suggested fix
In `docs/LANGUAGE.md` Control flow, add `break` / `continue` (mirroring SPEC §7 wording) and the `for { ... }` infinite-loop form, e.g. a bullet and a short example:

```go
for { v, ok := <-ch  if !ok { break }  println(str(v)) }   // bare loop + early exit
```

PR #338, opened 2026-07-04
fix(#90): clarify break/return is the only way out of an infinite for loop

### LINKED ISSUE #90: docs: language reference never documents that `break`/`continue` don't exist (no early loop exit)
## Problem

MFL is presented as "Go-flavored," and both the spec and the tour document the infinite loop form `for { ... }`:

- `SPEC.md` §7 Statements (lines ~183-186) lists `for cond { ... }` and `for { ... }` (infinite) as supported loops.
- `docs/LANGUAGE.md` §Control flow (lines ~172-194) documents `while`, `for cond`, and `for k, v := range`.

But the language has **no `break` and no `continue`** — they are not keywords (`lexer.go:27-32` keyword set: `func return if else while for true false nil var go type struct chan make map range arena`), and the parser has no handling for them. Typing `break` or `continue` is parsed as a bare identifier and fails later as an undefined reference. A reader is given an infinite `for { }` loop (used in the showcased servers, e.g. `framework/machweb.src:72-79` `serve`) with **no documented way to exit it except `return`**.

Neither `docs/LANGUAGE.md`, `SPEC.md`, nor `README.md` mentions this anywhere (grep for `break`/`continue` in those files returns nothing).

## Why it matters

The control-flow docs read as a complete enumeration of looping. A Go programmer will reflexively reach for `break`/`continue` and get a confusing "undefined" error with no hint that the feature simply doesn't exist. Anyone writing an accept/event loop needs to know the only exit is `return` (or never returning).

## Suggested fix

In `docs/LANGUAGE.md` §Control flow and `SPEC.md` §7, add an explicit note that `break` and `continue` are not supported, and that an infinite `for { ... }` is exited with `return` (or by structuring the loop condition). Alternatively, treat this as the tracking issue for *implementing* `break`/`continue` if that is preferred over documenting the limitation.

javimosch:

**Focus:** ISSUE FIX OBJECTIVE (GitHub Issue #90: "docs: language reference never documents that `break`/`continue` don't exist (no early loop exit)")

ISSUE DESCRIPTION:
## Problem

MFL is presented as "Go-flavored," and both the spec and the tour document the infinite loop form `for { ... }`:

- `SPEC.md` §7 Statements (lines ~183-186) lists `for cond { ... }` and `for { ... }` (infinite) as supported loops.
- `docs/LANGUAGE.md` §Control flow (lines ~172-194) documents `while`, `for cond`, and `for k, v := range`.

But the language has **no `break` and no `continue`** — they are not keywords (`lexer.go:27-32` keyword set: `func return if else while for true false nil var go type struct chan make map range arena`), and the parser has no handling for them. Typing `break` or `continue` is parsed as a bare identifier and fails later as an undefined reference. A reader is given an infinite `for { }` loop (used in the showcased servers, e.g. `framework/machweb.src:72-79` `serve`) with **no documented way to exit it except `return`**.

Neither `docs/LANGUAGE.md`, `SPEC.md`, nor `README.md` mentions this anywhere (grep for `break`/`continue` in those files returns nothing).

## Why it matters

The control-flow docs read as a complete enumeration of looping. A Go programmer will reflexively reach for `break`/`continue` and get a confusing "undefined" error with no hint that the feature simply doesn't exist. Anyone writing an accept/event loop needs to know the only exit is `return` (or never returning).

## Suggested fix

In `docs/LANGUAGE.md` §Control flow and `SPEC.md` §7, add an explicit note that `break` and `continue` are not supported, and that an infinite `for { ... }` is exited with `return` (or by structuring the loop condition). Alternatively, treat this as the tracking issue for *implementing* `break`/`continue` if that is preferred over documenting the limitation.

4. COMMIT referencing the issue: fix(#90): <brief description>
5. The PR body MUST include 'Fixes #90' so GitHub auto-closes the issue on merge.

**Branch:** `am/am-7b5889-thnec4`

## Summary

Issue #90 claimed break/continue don't exist in MFL, but the compiler already implements them (lexer.go keyword set, parser.go handling) and SPEC.md §7 already documented them. The one real residual gap — that break is the way to exit an infinite `for { ... }` loop — wasn't explicit, so I added a short clarifying clause to that bullet in SPEC.md. docs/LANGUAGE.md, the other file the issue named, is currently claimed by an in-flight PR (#321) so it was left untouched to avoid a conflict.

**Diff:**
```
SPEC.md | 3 ++-
 1 file changed, 2 insertions(+), 1 deletion(-)
```

### DIFF
--- a/SPEC.md
+++ b/SPEC.md
@@ -218,7 +218,8 @@ startup (before `main`; at `_initialize` for a wasm reactor).
 - `if cond { ... } else if cond { ... } else { ... }` — conditions are `bool`.
 - Loops:
   - `while cond { ... }`
-  - `for cond { ... }` and `for { ... }` (infinite)
+  - `for cond { ... }` and `for { ... }` (infinite) — exit an infinite loop
+    with `break` or `return`; there is no other way out
   - `for k, v := range x { ... }` over a slice (index, element), map (key,
     value), or string (index, 1-char). Either variable may be `_`.
   - `for v := range ch { ... }` over a channel: receives each value until the


javimosch:
⚠️ **Automaintainer intent check: OFF-OBJECTIVE**

FAIL: The diff adds "exit an infinite loop with `break` or `return`" — but the issue explicitly states `break`/`continue` are NOT implemented (not keywords, undefined identifier if used), so this ch

The verification gate passed, but this change may not address the stated objective. Review before merging.

javimosch:
**Review — Head of Org Engineering:** Small, correct doc clarification to SPEC.md about exiting infinite loops; touches only the intended file, is structurally valid Markdown, and has no correctness, security, or scope issues.

javimosch:
**Automaintainer Review Verdict:** CHANGES REQUESTED — The diff adds 'exit an infinite loop with `break` or `return`' but Issue #90 (the authoritative source for this PR) explicitly states that `break`/`continue` are **not** keywords in MFL — they are absent from the `lexer.go:27-32` keyword set and the parser has no handling for them. The repo owner's own 'OFF-OBJECTIVE' comment echoes this concern. If `break` is genuinely unimplemented, documenting it as a valid exit mechanism is actively misleading. The PR summary claims the issue was mistaken and that the compiler already implements `break`/`continue`, but provides no cited evidence (e.g. the specific `lexer.go`/`parser.go` lines). Before approving: please verify whether `break` is in the keyword set and the parser handles it; if yes, update the PR to cite the evidence; if no, the doc change should say `return` is the only exit from an infinite loop.

javimosch:
**Automaintainer Review Verdict: REQUEST_CHANGES**

The reviewer identified a correctness issue: this PR documents `break` as a valid exit from infinite loops, but Issue #90 explicitly states that `break`/`continue` are **not** MFL keywords. If the issue description is accurate, this change misleads users rather than fixing the documentation gap.

The diff needs verification against `lexer.go`'s keyword set before it can be approved. If `break` is indeed unsupported, the documentation should state that the **only** exit from an infinite `for {}` loop is `return` — not `break`.

javimosch:
Closing without merging — the underlying issue (#90) is stale (break/continue are fully implemented now, contradicting the issue's premise). See #90 for details. #150 tracks the real remaining doc gap.

Issue #313, opened 2026-07-03
Self-hosted codegen: port the #310 go-arg use-after-free fix (arena boundary deep-copy)


javimosch:
## Context

#312 (fixing #310) changed the reference (Go) compiler's C output for \`go f(args)\` when an argument is a string, or a struct/slice/map containing one — deep-copying the argument across the spawning goroutine's arena boundary, exactly like channel sends already do. The self-hosted compiler (\`selfhost/cgen.src\`) doesn't have this fix yet.

## Impact

\`selfhost/verify-cgen.sh\` now shows 3 corpus mismatches (was 411/0, now 408/3 PASS/FAIL) — confirmed by running it after #312 landed. One of the three is a genuine functional gap (a corpus file needing \`mfl_thaw_strs\` that the self-hosted output doesn't emit); the other two are cosmetic (a line-break placement difference in generated C, no functional difference for those particular calls since they only have scalar arguments).

## Suggested fix

Port the same change to \`selfhost/cgen.src\`'s go-statement codegen: classify each argument (scalar / string-bearing / slice-or-map-bearing) and emit the matching \`mfl_freeze_strs\`/\`mfl_thaw_strs\` calls or JSON round-trip, mirroring \`codegen.go\`'s \`goStmt\` (see #312's diff). The self-hosted compiler's own \`racecheck.src\` already has related dataflow logic for goroutine captures if any of that analysis is reusable.

## Not a user-facing blocker

Same framing as the other self-host completionist gaps (#293, #308): no real program uses \`selfhost/mfl-cgen\` as its primary compiler today; this is about keeping the self-hosting milestone's corpus parity from silently drifting.

javimosch:
Fixed and released in v0.105.0 — PR #315. Self-hosted compiler now emits byte-identical protection to the reference compiler; corpus parity restored (verify-cgen.sh 411/0), fixpoint and race-checker both confirmed unaffected.

Issue #310, opened 2026-07-03
[high] use-after-free: goroutine args captured from a machweb request arena are freed when the response is sent


javimosch:
**Priority: high** — memory-safety bug: an easy, natural pattern silently reads freed memory.

## Repro (pattern)

In a machweb handler, spawn background work after replying:

```go
func handle_chat(req) (res) {
    conv := "..."         // any string derived from the request / db reads
    go background_work(ag, conv)   // ag is a struct with string fields
    res = ok_json("{}")
}
```

machweb reclaims the request goroutine's arena as soon as the response is sent (as documented in machweb.src's header). The detached goroutine's arguments still point into that arena → use-after-free. Observed in production logs of machin-agents as corrupted strings:

```
{"event":"compacted","conv":"�%y","folded":2,"upto":16}
{"warn":"...","conv":"hat about time zone handling for remote teams?\"}�","detail":""}
```

(`conv` was a 16-hex-char id; the corrupted values are other freed request memory.)

## Suggested fixes (any of)

1. `go f(args...)` deep-copies string/struct/slice arguments into the new goroutine's arena (safest; matches user intuition from Go).
2. Escape analysis: values passed to `go` are promoted out of the arena.
3. At minimum: a loud warning in `machin guide` + machweb docs that `go` from a request handler must not capture request-arena values.

## Workaround used in machin-agents

Handlers persist pending work as flags in SQLite and spawn a **zero-argument** sweeper goroutine that re-reads everything from the db in its own arena (`needs_compact`/`needs_memory` columns + `sweep()`), with `changes()`-based claims for idempotency. Works, but nothing warns you before you hit the corruption.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

javimosch:
Fixed and released in v0.104.0 — PR #312. See CHANGELOG v0.104.0 for details; verified empirically (reproduced the actual corruption/segfault pre-fix, confirmed both gone post-fix) with two new regression tests.

PR #421, opened 2026-07-06
fix: complete #314 — closure-captured data now survives the go arena boundary

javimosch:
## Summary

- Completes #314 (reopened after verifying PR #376 was a partial fix — its own test passes, but the original repro still corrupted deterministically on current main).
- #376 made the closure env struct and capture boxes malloc-stable; this PR fixes the remaining level: the **data a box points at** (string bytes, struct string fields, slice/map backings) still lived in the spawning goroutine's arena. When the spawner is itself a short-lived goroutine — the repro's exact shape — that arena is reclaimed while the spawned closure still reads through the box.
- `goStmt` now detects a `*MakeClosure` argument with captures and freeze/thaws each capture's box contents across the arena boundary, reusing #310's machinery one indirection level deeper: string-offset copy for strings/structs-of-strings, a JSON round-trip side-field for slice/map-bearing captures, nothing for scalars.
- Semantics follow the channel-send doctrine: ownership of a captured variable moves at the `go` statement. A closure passed as a plain *variable* keeps SPEC.md §12's documented shared-value caveat (env layout is unknowable at the call site).

## Verification

- Original repro: **3/3 corrupted pre-fix → 5/5 clean post-fix** (same binary-diff discipline as #310/#312).
- 30-way stress across all three capture classes (string, struct-of-strings, `[]string`) with allocation-churn goroutines pressuring arena reuse: all values intact and distinct.
- New regression test `TestGoClosureCapturedDataSurvivesSpawnerArena` fails reliably (3/3) on pre-fix codegen, passes (5/5) on the fix.
- Full `go test -count=1 .` green.
- Self-hosted parity: `verify-cgen.sh` **412 PASS / 0 FAIL** — no #313-style port needed, since closures are `(unsupported)` in the self-hosted codegen and never enter the diff.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,23 @@
   arena-independent, matching how channels and maps already manage their own
   memory; string/slice/map captures reassigned through the box remain a
   separate, still-open gap. See #314.
+- **Completed the #314 fix: the DATA a captured variable holds now also
+  survives the arena boundary, not just the pointer chain.** The previous
+  entry made the env struct and capture boxes malloc-stable, but a captured
+  string's bytes, a captured struct's string fields, and a captured
+  slice/map's backing still lived in the spawning goroutine's arena — so
+  `go f(func(){ ...label... })` from a short-lived goroutine still read
+  corrupted memory once the spawner's arena was reclaimed (the original
+  repro was never actually fixed; verified before and after). `goStmt` now
+  classifies each capture of a closure literal passed to `go` and
+  freeze/thaws its box contents across the boundary — the same machinery
+  channel sends and plain `go` arguments already use (#310): string-offset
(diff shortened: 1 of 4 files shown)


javimosch:
**Review — Head of Org Engineering:** The diff implements the stated fix (freeze/thaw of captured closure DATA across the arena boundary in goStmt, complementing the earlier #310/#376 pointer-stability fix), adds a targeted regression test with corruption/duplicate detection, and only touches the four relevant files (codegen.go, types.go, CHANGELOG.md, and its own test file) with no stray edits or secrets. The generated C snippets and Fprintf format/argument counts are internally consistent, and the new VarTypeString mirrors the existing VarCType(fn, name) pattern, so no blocking correctness or scope issue is present.

PR #477, opened 2026-07-17
feat(replay): Phase 3e — value-query replay debugger (machin replay --print <var>)

javimosch:
## Phase 3e — value-query replay debugger

`machin replay <trace> --print <var>[,<var>...]` turns a recorded run into a debugger: it rebuilds the program with a probe after each assignment to the watched variable and re-runs the **exact recorded execution**, printing that variable's value history to stderr.

```
$ machin replay crash.mflrr --print idx
probe 0 idx = 0 @op0
probe 0 idx = 5 @op3
probe 0 idx = 12 @op4
panic: index out of range [12] with length 2
  (replay: goroutine 0, schedule op 4/4)
```

Because replay is deterministic, the sequence is **exactly** what the variable did in the recorded run — the last line before a panic is its value at the crash, and `@op<n>` correlates with the causal report's `scheduleOp`. You watch how the bad value built up.

### Implementation
- `cgen` gains `probeVars`; a package var `debugProbeVars` is set by `cmdReplay` around `BuildBinary` and cleared after — no build/compile signature churn, no leak into other builds.
- `emitProbe` runs after each `AssignStmt` (covers `:=` and `=`, one node with `.Op`) — scalars/strings only (`mfl_str_i/d/b`); slice/map/struct/chan skipped.
- runtime `mfl_rr_probe(name, val)` prints under `MFL_RR_PROBE`, tagged with goroutine path + schedule position.

### Zero cost off the debug path
Probes emit **only** when `--print` is given, so a normal `build`/`run` carries nothing. **No self-host port needed** — the `cgentest` oracle-diff never sets `probeVars`, so the Go and self-hosted compilers stay byte-identical (413/2); the prelude regen only adds the `mfl_rr_probe` symbol to CORE.

> `--at <line>` was infeasible: the AST carries no line numbers, only the parser's token positions. `--print <var>` is the chosen shape.

### Gates
- `verify-replay.sh` → **49/49** (+deterministic value history, ends at final value, normal run emits no probes)
- `selfhost/verify-cgen.sh` → **413 PASS / 2 pre-existing FAIL** (byte-identical)
- `go test .` → ok

Only the blog ("The crash you can mail") remains of the record/replay follow-ons.


javimosch:
❌ **automaintainer verify gate**: `failed`

<details><summary>Output</summary>

```
…01
machin: installed 1 skill(s) -> /tmp/TestSkillInstallSingleNamedSkill3009908531/001
machin: --static + TLS/crypto: linking a static OpenSSL (needs libssl-dev's static archives — .a files — on the build host; a plain `apt install libssl-dev` provides them, no musl needed for this part, unlike --static SQLite). Bundling a CA root store so the binary verifies certificates with no external files (FROM scratch).
2026/07/17 10:37:04 http: TLS handshake error from 127.0.0.1:37342: local error: tls: bad record MAC
0 root=0 kind=int
1 root=1 kind=func fsig=|0
err=0
0 root=0 kind=var
1 root=1 kind=num
2 root=2 kind=int
3 root=3 kind=float
4 root=4 kind=bool
5 root=5 kind=string
6 root=6 kind=void
7 root=7 kind=bytes
8 root=8 kind=slice elem=2
9 root=9 kind=chan elem=2
err=1
0 root=0 kind=int
1 root=1 kind=string
err=1
err=0
FAIL
FAIL	machin	183.381s
FAIL
```
</details>

PR #515, opened 2026-07-18
fix: int-literal 32-bit overflow + keys() use-after-free + test-helper collision

javimosch:
Three fixes surfaced in one day of building [grange](https://github.com/javimosch/grange) in MFL:

1. **Int literals emit as bare C constants** → `30 * 86400000` does 32-bit `int` arithmetic (UB overflow) before widening to `int64_t`. Now emitted with the `LL` suffix. Found when grange's billing accrual went negative.
2. **`keys(map)` use-after-free**: `mfl_map_keys` returned aliases of each entry's malloc'd key and `mfl_map_del` frees them — deleting while iterating a retained `keys()` snapshot reads freed memory. Keys now copy into arena storage. Found when a grange WAL tombstone id turned to garbage mid-sweep.
3. **`captureStdout` redeclared** in two test files (bad merge) — test compilation on main is currently broken; deduped.

Local verification: fixes confirmed by minimal repros + grange's 77-test suite; the full local `go test .` run wedges environmentally on this box (subprocess tests hang sequence-dependently regardless of these changes — verified by control runs on unmodified code), so CI is the gate here.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved null-string handling to prevent crashes and treat missing strings as empty.
  * Ensured map keys remain valid after map entries are removed.
  * Improved numeric code generation for consistent 64-bit arithmetic.

* **Tests**
  * Expanded coverage for command-line workflows, parsing, type checking, code generation, and build scenarios.
  * Added validation for parser and multi-assignment error handling.

* **Documentation**
  * Updated coverage requirements and documented upcoming physics-library integration and callback capabilities.


javimosch:
Superseded: equivalent fixes landed on main via #513 (64-bit int literals + keys() copy) and the captureStdout collision no longer exists there — released in v0.109.0. Closing.

javimosch:
Superseded. The two headline fixes (int-literal 32-bit overflow, keys() use-after-free) shipped in **v0.109.0** via #513. The NULL-string segfault fix was extracted onto current main in #516. This branch was based on pre-v0.109.0 main, so merging it would have reverted the arena-escape-analysis / superoptimization / release work; the Phase 1-6 coverage fixtures hang on CI against current main and were not carried over.

Issue #539, opened 2026-07-28
ARENA001: interprocedural gap lets a global dangle out of an arena — silent wrong data, no diagnostic


javimosch:
ARENA001 catches a global assigned from inside an `arena { }` block. It does not catch the same assignment **one function call deeper**, and the result is silent memory corruption in safe MFL.

This is not hypothetical: it segfaulted the [grange](https://github.com/javimosch/grange) server in production-shaped use. An ordered query built a lazily-built sorted projection (stored in a global) inside the request's scoped arena; the arena was freed at the end of the request, and the **second** query read freed memory. One query looked perfect, which is why it shipped.

## Repro

Caught today (warning):

```
var g_cache = []string{}

func main() {
    arena {
        out := []string{}
        i := 0
        while i < 200 { out = append(out, "entry-" + str(i))  i = i + 1 }
        g_cache = out          // <-- ARENA001 fires here
    }
    println("first=" + g_cache[0])
}
```

```
warning: [ARENA001] `g_cache` outlives this arena block but is assigned a value
allocated inside it — it dangles after the block's memory is reclaimed (line 3)
```

Not caught — identical semantics, assignment moved into a callee:

```
var g_cache = []string{}

func build_cache() {
    out := []string{}
    i := 0
    while i < 200 { out = append(out, "entry-" + str(i))  i = i + 1 }
    g_cache = out
}

func main() {
    arena { build_cache() }
    println("len=" + str(len(g_cache)))
    println("first=" + g_cache[0])
}
```

```
$ machin check repro.src
ok — no errors

$ ./repro
len=200
first=
exit=0
```

`len` still reads 200 while the element is gone: **silent wrong data, exit 0, no diagnostic**. In a larger program the same shape segfaults.

## Why this matters more than an ordinary false negative

ARENA001 is the evidence behind "sound bulk reclamation = memory safety without a borrow checker". A caller cannot currently tell, from any diagnostic, whether a function it calls inside `arena { }` stores an arena-allocated value in a global — and calling a helper inside an arena block is the normal way to write MFL. The analysis is sound for what it inspects, but the boundary is one stack frame deep, which is narrower than the claim reads.

Two things worth considering separately:

1. **Interprocedural provenance** — propagate "allocates into the current arena and stores it somewhere that outlives it" through calls. This is the real fix and a large one.
[...]

PR #545, opened 2026-08-03
docs(arena_reset): document the persistence rule for globals across a reset

### LINKED ISSUE #540: arena_reset() silently corrupts computed strings held in globals (literals/env/args survive)
Companion to #539, and the same underlying hazard from the other direction: #539 is about a value that escapes INTO a scoped `arena { }`; this is about what survives `arena_reset()`.

`arena_reset()` frees the arena, but which globals stay valid afterwards is not obvious and is not diagnosed. Measured:

| held in a global across `arena_reset()` | result |
|---|---|
| string literal | survives |
| `env("X")` value | survives |
| `args()` value | survives |
| **computed string** (concatenation, etc.) | **silently corrupted** |

```
var g_concat = ""

func main() {
    g_concat = "tok-" + env("RT_TOKEN") + "-end"
    println("before: " + g_concat)      // tok-secret123-end
    arena_reset()
    junk := []string{}
    i := 0
    while i < 20000 { junk = append(junk, "yyyy" + str(i))  i = i + 1 }
    println("after:  " + g_concat)      // 1
}
```

No crash, no diagnostic, exit 0 — the value is simply wrong. (Without the allocation churn it still *looks* correct, because the freed pages have not been reused yet, which makes casual testing actively misleading.)

## Why this is worth a diagnostic

`arena_reset()` is the documented way for a long-lived single-actor server to stay flat, so the programs most likely to call it are exactly the ones holding startup state in globals. It bit [grange](https://github.com/javimosch/grange) in a way that took a while to find:

```
var g_nl = bytes("\n")
```

One computed global — the separator every page, chunk and index walk searches for. After a reset, every line-walk searched for garbage: pages read as EMPTY and the integrity checker reported "malformed page records" on files that were perfectly fine. Counts stayed CORRECT throughout, because they answer from a manifest and never walk a page. So the failure presented as "the verifier is broken", not "memory is corrupt".

The same shape hits any consumer: a db path, an auth token, a cached prefix — anything built with `+` at startup and kept in a global.

## What would help, roughly in order of value

1. **A diagnostic.** At an `arena_reset()` call, flag globals whose current value has arena provenance — the same provenance machinery #539 wants, applied at the reset point rather than at a block exit. Even a conservative over-approximation ("this global may hold arena-allocated data at reset") would have turned a day of bisecting into a compile-time note.
2. **Documentation of the rule.** `machin guide` describes `arena_reset()` without stating what survives it. The literal/env/args-vs-computed distinction is learnable but currently only by experiment.
3. **A way to opt a value out** — an `intern()`/`persist()` that allocates outside the resettable arena, so a library can hold a computed string safely. Today the only re-derivation sources are `args()`/`env()`/literals, which is why grange re-parses argv after each reset instead of reading its own parsed flag struct (whose maps are themselves computed and therefore also corrupted).

Consumer-side workaround, for reference: grange re-derives its db path and token from `args()`/`env()` directly after every reset, refuses to enable periodic reset at all when the token was randomly generated (a computed string that cannot be re-derived), and re-initialises `g_nl` inside its own reset wrapper so embedders inherit the fix.

javimosch:
## Summary
- Documents option 2 from #540: the literal/env/args-vs-computed distinction for globals held across `arena_reset()`.
- **guide.go** (`machin guide` arena_reset entry): adds a PERSISTENCE RULE note — string literals, `env()` values, and `args()` values survive `arena_reset()` (backing memory not on the arena chain), but computed strings (concatenation, etc.) held in globals are silently corrupted (freed pages may be reused, no crash, no diagnostic). Re-derive computed globals from args()/env()/literals after each reset.
- **SPEC.md** (arena_reset section): adds the full persistence table (literal/env/args survive vs computed corrupted), the grange incident as a concrete example, and the consumer-side workaround (re-derive after each reset, or keep cross-reset state in malloc-backed maps/channels or on disk).

This is a docs-only change (option 2 from the issue). The diagnostic (option 1) and `intern()`/`persist()` builtin (option 3) are left for separate HITL work.

Closes #540

#### Test plan
- [x] `go build ./...` passes (string-literal-only change to the guide catalog)
- [ ] Human review: confirm the persistence rule wording matches the measured behavior in the issue body

Generated with [Devin](https://devin.ai)

## Summary by CodeRabbit

* **Documentation**
  * Clarified how `arena_reset()` affects string data.
  * Documented which values persist across resets and warned that computed global strings may become corrupted.
  * Added guidance to re-derive computed values or store persistent state outside arena memory.


javimosch:
## Review — merge, but **#540 should stay open**.

The entry is accurate and well-argued: it names the mechanism (literals/`env`/`args` are
re-derivation sources off the arena chain; a computed string is arena-allocated), gives the
workaround, and cites the real `grange` incident. Documentation this specific is worth having.

But it documents a **silent memory-corruption bug**, it does not fix one. Re-verified on
v0.124.1 just now:

```
before: tok-secret123-end
after : 1
$ machin check repro.src
ok — no errors
```

A global reads back as `1`, exit 0, and the checker is clean. That is the same shape as #539 —
an arena hazard the analysis cannot see — and the same argument applies: ARENA001 exists
precisely so that "sound bulk reclamation = memory safety without a borrow checker" is a claim
the compiler backs, and a documented footgun does not back it.

Merging this improves the situation today. Keeping #540 open is what keeps the real fix on the
table: either extend the escape analysis to globals held across `arena_reset()`, or make
`arena_reset()` refuse to run while a computed global points into the arena.

Issue #549, opened 2026-08-04
windows target broken on main: emitted runtime calls undeclared 'fsync' (zig cc, C11)


javimosch:
## `main` cannot cross-compile to Windows

`TestWindowsNetCompiles` and `TestWindowsBuildsPE` both fail on `main` at `565c25c`:

```
$ go test -run 'TestWindowsNetCompiles|TestWindowsBuildsPE' ./...
windows_target_test.go:120: BuildWindows (net): zig failed: exit status 1
  /tmp/mfl-817508131.c:1741:13: error: call to undeclared function 'fsync';
  ISO C99 and later do not support implicit function declarations [-Wimplicit-function-declaration]
  1 error generated.
--- FAIL: TestWindowsNetCompiles (0.91s)
--- FAIL: TestWindowsBuildsPE (0.88s)
```

The emitted runtime calls `fsync(...)` unconditionally, but the Windows target has no
declaration for it — `fsync` is POSIX and mingw does not provide it. `zig cc` for
`x86_64-windows-gnu` compiles as C11, where an implicit declaration is a hard error.

## Origin

The `fsync` builtin landed in v0.122.0 (#536 / #538) without a Windows path. The Windows
target already carries this kind of guard elsewhere (`build.go`'s comment about winsock2
and winpthreads for #517), so the pattern to follow exists.

## Impact

`--target windows` is broken on `main` — not a flake. It also puts two permanent reds in
`go test ./...`, which means every AM run on this repo comes back `verify_failed` for a
reason unrelated to whatever it changed. That is how a real failure gets ignored.

## Suggested fix

In the runtime C preamble, guard the POSIX call and map it to the Win32 equivalent:

```c
#ifdef _WIN32
  #include <io.h>
  static int mfl_fsync(int fd) { return _commit(fd); }
#else
  static int mfl_fsync(int fd) { return fsync(fd); }
#endif
```

then emit `mfl_fsync(...)` instead of `fsync(...)`. `_commit` is the documented MSVCRT
equivalent (flush a file descriptor's buffers to disk) and is what mingw programs use.

## Acceptance criteria

1. `go test -run 'TestWindowsNetCompiles|TestWindowsBuildsPE' ./...` passes on Linux with
   `zig` installed.
2. `fsync` on Windows actually commits — do not stub it to a no-op returning success. If a
   true equivalent is unavailable for some fd kind, fail loudly rather than silently
   claiming durability, since that is the entire point of the builtin.
[...]

PR #559, opened 2026-08-05
fix(#552): panic on narrow cstruct field overflow under --safe

### LINKED ISSUE #552: u8/u16 cstruct fields narrow silently at the FFI boundary — MFL and C hold different values
## Problem

An out-of-range value assigned to a `u8` (or `u16`) `cstruct` field is silently truncated
**at the C boundary only**. The MFL-side value keeps the full integer, so MFL and C hold
*different values* for the same field and the program cannot detect it by reading the field
back.

### Repro (raylib's `Color`, but any narrow cstruct field does it)

```
base := Color{240, 120, 40, 255}
f := 1.5
scaled := Color{int(float(base.r) * f), int(float(base.g) * f), int(float(base.b) * f), 255}
println("out: r=" + str(scaled.r))
```

```
out: r=360
```

But the emitted C narrows on the way out:

```c
typedef struct { int64_t f_r; int64_t f_g; int64_t f_b; int64_t f_a; } mfl_Color;
static Color mfl_to_Color(mfl_Color m) {
    return (Color){ .r = (uint8_t)m.f_r, .g = (uint8_t)m.f_g, ... };
}
```

`(uint8_t)360` is `104`. So `DrawRectangle(..., scaled)` draws with **r=104**, while
`str(scaled.r)` in the same program says **360**. There is no diagnostic on either side.

## How it actually bit

Tinting a sprite by 1.5× turned bright orange fire **green** — `r` wrapped past 255 and
came back low while `g` and `b` did not. It survived a code review because the MFL code
reads correctly and printing the field shows the value you expect. It was only findable by
looking at the rendered pixels.

## Suggested fix

Three options, cheapest first — any one of them removes the silence:

1. **Compile-time**: when a `cstruct` field of a narrow type is assigned a *constant*
   expression that provably exceeds the field's range, error.
2. **`--safe`**: emit a checked narrowing in `mfl_to_<T>` that panics
   (`integer overflow (u8 field 'r')`), consistent with `--safe`'s existing
   `__builtin_*_overflow` treatment of `+ - *`.
3. **Documentation** (minimum): a `guide.go` gotcha stating that narrow cstruct fields
   truncate at the FFI boundary and that reading the field back in MFL will **not** reveal
   it.

(2) plus (3) is the honest combination: defined-and-observable under `--safe`, documented
everywhere else.

## Acceptance criteria

1. A test asserting the current divergence is either caught or documented — i.e. under
   `--safe`, the repro panics rather than drawing the wrong colour.
2. If (1) is implemented, a `guide.go` gotcha describing the behaviour in both modes.
3. Applies to every narrow cstruct field type in use (`u8`, `u16`, `i32` where the MFL
   value is `int64`), not only `u8`.

## Out of scope

Do not change the default (non-`--safe`) narrowing behaviour to panic — that would be a
performance and compatibility change on every FFI call. Silent-but-documented is
acceptable for the fast path; silent-and-undocumented is not.

javimosch:

**Focus:** ISSUE FIX OBJECTIVE (GitHub Issue #552: "u8/u16 cstruct fields narrow silently at the FFI boundary — MFL and C hold different values")

ISSUE DESCRIPTION:
## Problem

An out-of-range value assigned to a `u8` (or `u16`) `cstruct` field is silently truncated
**at the C boundary only**. The MFL-side value keeps the full integer, so MFL and C hold
*different values* for the same field and the program cannot detect it by reading the field
back.

### Repro (raylib's `Color`, but any narrow cstruct field does it)

```
base := Color{240, 120, 40, 255}
f := 1.5
scaled := Color{int(float(base.r) * f), int(float(base.g) * f), int(float(base.b) * f), 255}
println("out: r=" + str(scaled.r))
```

```
out: r=360
```

But the emitted C narrows on the way out:

```c
typedef struct { int64_t f_r; int64_t f_g; int64_t f_b; int64_t f_a; } mfl_Color;
static Color mfl_to_Color(mfl_Color m) {
    return (Color){ .r = (uint8_t)m.f_r, .g = (uint8_t)m.f_g, ... };
}
```

`(uint8_t)360` is `104`. So `DrawRectangle(..., scaled)` draws with **r=104**, while
`str(scaled.r)` in the same program says **360**. There is no diagnostic on either side.

## How it actually bit

Tinting a sprite by 1.5× turned bright orange fire **green** — `r` wrapped past 255 and
came back low while `g` and `b` did not. It survived a code review because the MFL code
reads correctly and printing the field shows the value you expect. It was only findable by
looking at the rendered pixels.

## Suggested fix

Three options, cheapest first — any one of them removes the silence:

1. **Compile-time**: when a `cstruct` field of a narrow type is assigned a *constant*
   expression that provably exceeds the field's range, error.
2. **`--safe`**: emit a checked narrowing in `mfl_to_<T>` that panics
   (`integer overflow (u8 field 'r')`), consistent with `--safe`'s existing
   `__builtin_*_overflow` treatment of `+ - *`.
3. **Documentation** (minimum): a `guide.go` gotcha stating that narrow cstruct fields
   truncate at the FFI boundary and that reading the field back in MFL will **not** reveal
   it.

(2) plus (3) is the honest combination: defined-and-observable under `--safe`, documented
everywhere else.

## Acceptance criteria

1. A test asserting the current divergence is either caught or documented — i.e. under
   `--safe`, the repro panics rather than drawing the wrong colour.
2. If (1) is implemented, a `guide.go` gotcha describing the behaviour in both modes.
3. Applies to every narrow cstruct field type in use (`u8`, `u16`, `i32` where the MFL
   value is `int64`), not only `u8`.

## Out of scope

Do not change the default (non-`--safe`) narrowing behaviour to panic — that would be a
performance and compatibility change on every FFI call. Silent-but-documented is
acceptable for the fast path; silent-and-undocumented is not.

4. COMMIT referencing the issue: fix(#552): <brief description>
5. The PR body MUST include 'Fixes #552' so GitHub auto-closes the issue on merge.

**Branch:** `am/am-84a3be-dkhautnrnfej-c78f7506`

```
  searching for a cheaper equivalent of x...
  no cheaper equivalent found within the search bounds  (explored 0 candidates, 0 distinct behaviors)
0 root=0 kind=int
1 root=1 kind=func fsig=|0
err=0
0 root=0 kind=var
1 root=1 kind=num
2 root=2 kind=int
3 root=3 kind=float
4 root=4 kind=bool
5 root=5 kind=string
6 root=6 kind=void
7 root=7 kind=bytes
8 root=8 kind=slice elem=2
9 root=9 kind=chan elem=2
err=1
0 root=0 kind=int
1 root=1 kind=string
err=1
err=0
--- FAIL: TestWindowsNetCompiles (0.65s)
        /tmp/mfl-933069463.c:1745:13: error: call to undeclared function 'fsync'; ISO C99 and later do not support implicit function declarations [-Wimplicit-function-declaration]
         1745 |     int r = fsync(fd);
              |             ^
        1 error generated.
--- FAIL: TestWindowsBuildsPE (0.69s)
        /tmp/mfl-3465161629.c:1745:13: error: call to undeclared function 'fsync'; ISO C99 and later do not support implicit function declarations [-Wimplicit-function-declaration]
         1745 |     int r = fsync(fd);
              |             ^
        1 error generated.
FAIL
FAIL	machin	160.330s
FAIL
```

## Summary

Fixed #552: u8/u16/i32 cstruct fields no longer silently diverge between MFL and C. Under --safe, mfl_to_<Struct> now range-checks each narrow field and panics with "integer overflow (<ctype> field '<name>')" instead of wrapping, matching --safe's existing +/-/* overflow behavior. The default (non-safe) fast path is unchanged and now documented via a new guide.go gotcha explaining the truncation happens only at the FFI boundary and is invisible when reading the field back in MFL. Added TestNarrowCStructFieldSafe covering both build modes.

**Diff:**
```
codegen.go  | 27 +++++++++++++++++++++++++++
 guide.go    |  1 +
 mfl_test.go | 38 ++++++++++++++++++++++++++++++++++++++
 3 files changed, 66 insertions(+)
```


javimosch:
**Automaintainer Review Verdict:** APPROVED — Fix is correct and well-scoped. `mfl_narrow()` C helper + `narrowRange()` Go function cover all narrow cstruct field types (i8/i16/i32/u8/u16/u32). The `--safe` gate in codegen is correct, non-safe fast path is unchanged, guide.go gotcha is accurate, and `TestNarrowCStructFieldSafe` covers both build modes via emitted-C inspection. The failing CI tests (`TestWindowsNetCompiles`, `TestWindowsBuildsPE`) are pre-existing Windows cross-compile failures from the `fsync` builtin added in #538 — unrelated to this PR (0 deletions, no Windows code touched). Addresses #552 as specified.

javimosch:

javimosch:
## Architect review — **the best of the batch so far. Merge it.**

Every acceptance criterion met, one of them exceeded, and the "out of scope" boundary
respected exactly.

### Verified behaviourally, not just by reading the codegen

The PR's own test asserts the *emitted C* contains (or lacks) the check. That is a
reasonable proxy but it is not proof the panic fires, so I built the real thing against
raylib's `Color`:

```
$ machin build u8safe.mfl -o u8s --safe && ./u8s
panic: integer overflow (u8 field 'r')
```

That is the exact scenario from the issue — `int(float(c.r) * 1.5)` overflowing a `u8` on
its way to `DrawRectangle` — and it now stops instead of drawing the wrong colour. The
label names the field, so you get `r`, not just "somewhere in this struct".

### Criteria

| # | criterion | verdict |
|---|---|---|
| 1 | under `--safe`, the repro panics | ✅ verified above, exact label |
| 2 | `guide.go` gotcha | ✅ and it is unusually good — see below |
| 3 | applies to every narrow field type, not only `u8` | ✅ **exceeded**: `i8 i16 i32 u8 u16 u32` |
| — | out of scope: *do not* change the default path | ✅ gated on `g.safe`; default build emits **0** narrow calls and still plain-casts |

**On (2)** — the entry states the part that actually bites: that the MFL-side value keeps the
full `int64`, so *reading the field back does not reveal the divergence*. That is the reason
this survived a code review in the first place, and most write-ups would have omitted it. It
also says why the default stays unchecked (per-call cost) rather than leaving it looking like
an oversight.

**On (3)** — I asked for "not only u8" and got the full narrow set with a clean
`narrowRange` lookup. `i64`/`u64` are correctly absent: they are not narrower than MFL's int.

### Two nits, neither blocking

1. **`mfl_narrow` is emitted unconditionally**, so a default build carries an unused static
   and one `-Wunused-function`. In context this is nothing: the same build already has
   **149** such warnings from the `mfl_to_*`/`mfl_from_*` converters, and `machin build` does
   not pass `-Wall`. Only worth gating if someone ever cleans up that whole class.
2. The new test pins the emitted C rather than the runtime behaviour. Given the extern needs
   a real C library, that is a defensible trade — but the repro I ran above would make a
   stronger regression test if there is an existing pattern for building against a vendored
   lib in the suite.

### `verify_failed` is not this PR

`TestWindowsNetCompiles` / `TestWindowsBuildsPE`, the same pre-existing
`call to undeclared function 'fsync'` — #549. That is now **three of three** machin PRs
reported `verify_failed` for a reason unrelated to their change. Worth fixing #549 early,
purely so the gate starts carrying signal again.

### Recommendation

**Merge**, drop the `[needs work]` prefix. This one can close #552 — the issue's asks are
done. Leaving it open for the owner.

PR #548, opened 2026-08-04
fix(build): pass -fwrapv to all three cc invocations

javimosch:
Automated maintenance run by automaintainer.

**Focus:** ISSUE FIX OBJECTIVE (GitHub Issue #547: "codegen: pass -fwrapv — signed overflow is UB today and folds to a constant at -O2")

ISSUE DESCRIPTION:
## Problem

`machin build` compiles the emitted C with `-O2` but does **not** pass `-fwrapv`, so
signed integer overflow is **undefined behaviour** and the C optimizer is free to fold it.
That is not theoretical — it silently produces wrong arithmetic in release builds only.

Minimal MFL repro (a textbook 64-bit FNV-1a round):

```
func fnv64(h, v) (o) {
    o = h
    x := v
    i := 0
    while i < 8 {
        o = (o ^ (x & 0xFF)) * 0x100000001B3
        x = x >> 8
        i = i + 1
    }
}
func main() {
    h := 0x4BF29CE484222325
    println(str(fnv64(h, 0)) + " " + str(fnv64(h, 4242)) + " " + str(fnv64(fnv64(h, 0), 777)))
}
```

Measured on this repo's emitted C (gcc 13, `-std=c11 -fno-strict-aliasing`):

| build | `fnv64(h,0)` | `fnv64(h,4242)` | `fnv64(fnv64(h,0),777)` | correct? |
|---|---|---|---|---|
| **`-O2` (what `machin build` does today)** | `9223372036854775807` | `9125914303777821415` | `9223372036854775807` | ✗ folded to `INT64_MAX` |
| `-O2 -fwrapv` | `2938590176187398597` | `9125914303777821415` | `3527166882971247581` | ✓ |
| `-O0` | `2938590176187398597` | `9125914303777821415` | `3527166882971247581` | ✓ |

The `-fwrapv` / `-O0` values are exactly the two's-complement results (verified against an
independent 64-bit-masked implementation). Under `-O2` the multiply chain collapses to a
constant, so **any hash or PRNG built in MFL silently stops discriminating** — a state
checksum returned the same digest for two different worlds, with no error anywhere.

## Why this matters beyond hashing

- MFL is described as Go-flavored, and **Go's spec mandates two's-complement wrapping** for
  signed integers. Today MFL's integer semantics are `-O0`-vs-`-O2` dependent, i.e. defined
  by the backend rather than by the language.
- `certify`, `equiv`, `optimize` and `superopt` all reason about integer arithmetic that the
[...]

PR #560, opened 2026-08-05
fix(#553): support non-empty []struct literals

### LINKED ISSUE #553: support non-empty []struct literals
## Problem

```
type P struct { a int  b int }
func main() { xs := []P{P{1,2}, P{3,4}}  println(str(len(xs))) }
```

```
error: non-empty []struct literals are not supported; build with append
```

The diagnostic is clear and tells you the workaround — this issue is not about the message,
it is a request to **support the literal**.

## Why this one

It is the single most-cited caveat across every machin skill and demo repo. The workaround
turns a declaration into a paragraph:

```
xs := []P{}
xs = append(xs, P{1, 2})
xs = append(xs, P{3, 4})
```

Table-driven code is where it hurts most — material tables, tile legends, level rows,
test fixtures — exactly the data-as-code shape MFL is otherwise good at. `[]int{1,2,3}`
and `[]string{"a","b"}` already work, so the asymmetry is the surprise.

## Suggested fix

Support element literals in a `[]T` composite where `T` is a user struct, lowering to the
same sequence the workaround writes by hand (a zeroed slice of the right length, then a
field-wise store per element). Nested struct literals inside the elements should work
(`[]P{P{1,2}}` and, if the shorthand is supported elsewhere, `[]P{{1,2}}`).

## Acceptance criteria

1. `xs := []P{P{1,2}, P{3,4}}` compiles; `len(xs) == 2` and both elements read back correctly.
2. Works for a struct containing a slice or a string field, not only scalars.
3. Nested: `[]P{P{1,2}}` where `P` has a struct field.
4. The existing error message is removed only once the feature works — do not leave a
   half-supported path that accepts the syntax and miscompiles.
5. A test in the same style as the existing `structlit_test.go`.

## Out of scope

Map literals with struct values, and `[]T` where `T` is a cstruct (cstructs cannot be MFL
struct fields today; that is a separate limitation).

javimosch:
Automated run for issue #553. The run pushed this commit but PR creation failed on a transient GitHub API error (`error_during_execution` after 30 turns; the same token opened #548/#558/#559 minutes earlier). Opened manually so the work isn't lost.

Closes #553 once reviewed.

## Summary by CodeRabbit

- **New Features**
  - Added support for non-empty struct slice literals, including nested structs and fields containing slices or strings.
  - Generated slices can be compiled, accessed, and evaluated as expected.

- **Bug Fixes**
  - Resolved failures when using non-empty struct slices in supported literal expressions.

- **Tests**
  - Added regression coverage for compilation, element access, lengths, and output.

### DIFF
--- a/codegen.go
+++ b/codegen.go
@@ -5519,8 +5519,10 @@ func (g *cgen) sliceLit(ex *SliceLit) (string, error) {
 		builder, cast = "mfl_lit_str", "char*"
 	case KInt, KBool:
 		builder, cast = "mfl_lit_i64", "int64_t"
+	case KStruct:
+		return g.structSliceLit(ex)
 	default:
-		// struct / nested-slice elements: build empty + append in source instead
+		// nested-slice elements: build empty + append in source instead
 		return "", fmt.Errorf("non-empty []%s literals are not supported; build with append", ek)
 	}
 	return g.seqExprs(ex.Elems, func(names []string) (string, error) {
@@ -5530,6 +5532,32 @@ func (g *cgen) sliceLit(ex *SliceLit) (string, error) {
 		}
 		return builder + "(" + strings.Join(parts, ", ") + ")", nil
 	})
+}
+
+// structSliceLit builds a non-empty []T literal whose element T is a struct:
+// an arena-backed T array sized to the literal, filled one element per
+// assignment (the same shape the append-based workaround writes by hand),
+// wrapped in an mfl_slice.
+func (g *cgen) structSliceLit(ex *SliceLit) (string, error) {
+	ct := g.c.ElemCType(g.curFn, ex)
+	n := len(ex.Elems)
+	names := make([]string, n)
+	for i, el := range ex.Elems {
+		s, err := g.expr(el)
+		if err != nil {
+			return "", err
+		}
+		names[i] = s
+	}
+	id := g.tmpID
+	g.tmpID++
+	arr := fmt.Sprintf("_arrlit%d", id)
+	var b strings.Builder
+	fmt.Fprintf(&b, "%s* %s = (%s*)mfl_alloc(%d * sizeof(%s)); ", ct, arr, ct, n, ct)
+	for i, s := range names {
+		fmt.Fprintf(&b, "%s[%d] = %s; ", arr, i, s)
+	}
+	return fmt.Sprintf("({ %s(mfl_slice){%s, %d, %d}; })", b.String(), arr, n, n), nil
 }
 
 // stringZeroInits returns designated initializers that set every string reachable in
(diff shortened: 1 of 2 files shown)


javimosch:
## Architect review — **correct and complete. Merge.**

This PR had to be opened by hand: the run committed and pushed, then ended
`error_during_execution (30 turns)` before the publish step opened a PR. The run note
suggests the GitHub token lacks write access — **that diagnosis is wrong**: the same token
opened #548, #558 and #559 within the hour. The agent ran out of turns. Worth fixing the
note, because it points at the wrong thing.

### Criteria

| # | criterion | verdict |
|---|---|---|
| 1 | `[]P{P{1,2}, P{3,4}}` compiles, `len == 2`, elements read back | ✅ |
| 2 | works for a struct with a slice or string field | ✅ `Row{name string, vals []int}` |
| 3 | nested struct literal inside an element | ✅ `Inner{tag: "t0"}` |
| 4 | old error removed only once the feature works | ✅ `KStruct` case added; the `default` still rejects other element kinds |
| 5 | test in the existing `structlit_test.go` style | ✅ |

### Edges I probed beyond the tests

```
basic len=2 14          # []P{P{1,2}, P{3,4}}
empty len=0             # []P{} still works
after append len=3 last=56   # append onto a literal-built slice
rows=r02r11             # string + []int fields
single=19               # one-element literal
```

The one that mattered: **append onto a literal-built slice**. `structSliceLit` returns
`{arr, n, n}` with `len == cap`, so the first append must reallocate through `mfl_append`
rather than write past the arena block. It does — `len=3` with correct values.

### The real risk, and it is handled

`structSliceLit` allocates with `mfl_alloc`, i.e. from the **current arena**. A literal built
inside `arena { }` and stored somewhere outliving it would dangle. It is caught:

```
$ machin check ar.src
warning: [ARENA001] `g_rows` outlives this arena block but is assigned a value
allocated inside it — it dangles after the block's memory is reclaimed (line 3)
```

So the new allocation site is covered by the existing escape analysis and introduces no new
unsafety. That was the thing worth checking on this change and it comes out clean.

### Suite

Three failures, none of them this PR:

- `TestWindowsNetCompiles` / `TestWindowsBuildsPE` — the pre-existing `fsync` red, #549.
- `TestMachwebMultipartUpload` — **a flake**. It fails under full-suite load and **passes on
  both `main` and this branch when run alone**. Worth its own issue if it recurs; it is not a
  regression here.

### Nit

The `default:` arm's message now reads "nested-slice elements: build empty + append", but
`[][]int{...}` is rejected earlier by the **parser** (`slice element type must be a name, got "["`),
so that wording describes a case that can no longer reach it. Harmless, slightly misleading
if someone hits the arm via a different element kind.

### Recommendation

**Merge** — closes #553. This removes the single most-cited caveat across the machin skills
and demo repos.

PR #561, opened 2026-08-05
fix(#555): accept `_` as the sole assignment destination

### LINKED ISSUE #555: '_' is not assignable as a single destination (_ = f() errors) though it works in multi-assign
## Problem

`_` works as one destination inside a multi-assign, but not as the only destination.

```
func f() (v) { v = 1 }
func main() { _ = f()  println("ok") }
```

```
error: typecheck: assignment to undefined variable "_"
```

Meanwhile this is fine today:

```
s, _, _ = sim_spawn(s, e)
```

So `_` is already understood as "discard" in one position and not in another. The
inconsistency is the bug — you hit it the moment you want to call something for its side
effect and ignore its single return.

## Workaround

Bind a throwaway name, which then has to be *used* or it may itself warn:

```
unused := f()
```

## Suggested fix

Treat `_` as the blank identifier everywhere an assignment destination is accepted:
`_ = f()`, `_ := f()`, and inside multi-assign (already works). It is never readable, never
allocated, and never counts as a use.

## Acceptance criteria

1. `_ = f()` compiles and discards the value.
2. `_ := f()` behaves the same (or is rejected with a message pointing at `_ =`, if `:=`
   with only a blank is deliberately disallowed — either is fine, but pick one and say so).
3. Reading `_` remains an error (`x := _` must not compile).
4. `_` as a *function parameter* name and as a `range` variable behave consistently with
   whatever is decided; state the rule in the guide if it changes.
5. A test alongside the existing assignment tests.

## Out of scope

Unused-variable warnings generally.

javimosch:

**Focus:** ISSUE FIX OBJECTIVE (GitHub Issue #555: "'_' is not assignable as a single destination (_ = f() errors) though it works in multi-assign")

ISSUE DESCRIPTION:
## Problem

`_` works as one destination inside a multi-assign, but not as the only destination.

```
func f() (v) { v = 1 }
func main() { _ = f()  println("ok") }
```

```
error: typecheck: assignment to undefined variable "_"
```

Meanwhile this is fine today:

```
s, _, _ = sim_spawn(s, e)
```

So `_` is already understood as "discard" in one position and not in another. The
inconsistency is the bug — you hit it the moment you want to call something for its side
effect and ignore its single return.

## Workaround

Bind a throwaway name, which then has to be *used* or it may itself warn:

```
unused := f()
```

## Suggested fix

Treat `_` as the blank identifier everywhere an assignment destination is accepted:
`_ = f()`, `_ := f()`, and inside multi-assign (already works). It is never readable, never
allocated, and never counts as a use.

## Acceptance criteria

1. `_ = f()` compiles and discards the value.
2. `_ := f()` behaves the same (or is rejected with a message pointing at `_ =`, if `:=`
   with only a blank is deliberately disallowed — either is fine, but pick one and say so).
3. Reading `_` remains an error (`x := _` must not compile).
4. `_` as a *function parameter* name and as a `range` variable behave consistently with
   whatever is decided; state the rule in the guide if it changes.
5. A test alongside the existing assignment tests.

## Out of scope

Unused-variable warnings generally.

4. COMMIT referencing the issue: fix(#555): <brief description>
5. The PR body MUST include 'Fixes #555' so GitHub auto-closes the issue on merge.

**Branch:** `am/am-84a3be-dkhcdy8wpoc0-f54cb599`

## Summary

Fixes #555: `_` is now accepted as the sole destination of an assignment, not just inside a multi-assign. `_ = f()` and `_ := f()` both discard the value (evaluating it for side effects but never allocating a local for it), matching the existing multi-assign behavior. Reading `_` back remains a compile-time error. Patched all three layers involved (typecheck in types.go, codegen.go, and transform.go's declared-locals collector) plus a SPEC.md note and 3 new tests in assign_undefined_test.go.

**Diff:**
```
SPEC.md                  |  5 +++++
 assign_undefined_test.go | 47 +++++++++++++++++++++++++++++++++++++++++++++++
 codegen.go               |  6 ++++++
 transform.go             |  2 +-
 types.go                 |  7 +++++++
 5 files changed, 66 insertions(+), 1 deletion(-)
```

### DIFF
--- a/SPEC.md
+++ b/SPEC.md
@@ -394,6 +394,11 @@ A multi-value call may appear only as the sole right-hand side of a
 multi-assignment. `_` discards a value. A function returning ≥2 values compiles
 to a generated result struct.
 
+`_` is also accepted as the sole destination of a single-value assignment —
+`_ = f()` and `_ := f()` both discard the result, identically to `_` inside a
+multi-assign. `_` is write-only in every position: it is never allocated, never
+counts as a use, and reading it back (`x := _`) is a compile-time error.
+
 **Named returns.** A function may name its return values in the signature; they
 become zero-initialized locals, and a bare `return` (or falling off the end)
 yields their current values:
--- a/assign_undefined_test.go
+++ b/assign_undefined_test.go
@@ -84,3 +84,50 @@ func TestAssignFlatScopeReassignOK(t *testing.T) {
 		t.Fatalf("check: %v", err)
 	}
 }
+
+// TestAssignBlankSingleDestOK confirms `_` is accepted as the sole assignment
+// destination (not just inside a multi-assign): the value is discarded, never
+// allocated as a local, and the assignment is never treated as undefined.
+func TestAssignBlankSingleDestOK(t *testing.T) {
+	prog, err := ParseProgram([]string{
+		`func f() (v) { v = 1 }`,
+		`func main() { _ = f() println("ok") }`,
+	})
+	if err != nil {
+		t.Fatalf("parse: %v", err)
+	}
+	if _, err := Check(prog); err != nil {
+		t.Fatalf("check: %v", err)
+	}
+}
+
+// TestAssignBlankSingleDestWalrusOK confirms `_ := f()` behaves the same as
+// `_ = f()`: the blank identifier discards the value either way.
+func TestAssignBlankSingleDestWalrusOK(t *testing.T) {
+	prog, err := ParseProgram([]string{
+		`func f() (v) { v = 1 }`,
+		`func main() { _ := f() println("ok") }`,
+	})
+	if err != nil {
+		t.Fatalf("parse: %v", err)
+	}
+	if _, err := Check(prog); err != nil {
+		t.Fatalf("check: %v", err)
+	}
+}
+
+// TestAssignBlankUnreadableRejected confirms `_` remains write-only: reading
+// it back (`x := _`) must still be a compile-time error.
+func TestAssignBlankUnreadableRejected(t *testing.T) {
+	prog, err := ParseProgram([]string{
+		`func f() (v) { v = 1 }`,
+		`func main() { _ = f() x := _ println(str(x)) }`,
+	})
+	if err != nil {
+		t.Fatalf("parse: %v", err)
+	}
+	_, err = Check(prog)
+	if err == nil || !strings.Contains(err.Error(), `undefined variable "_"`) {
+		t.Fatalf("expected undefined-variable error for reading _, got %v", err)
+	}
+}
--- a/codegen.go
+++ b/codegen.go
@@ -4474,6 +4474,12 @@ func (g *cgen) stmt(s Stmt, depth int) error {
 		if err != nil {
 			return err
 		}
+		if st.Name == "_" {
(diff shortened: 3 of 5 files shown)


javimosch:
⚠️ **Automaintainer intent check: OFF-OBJECTIVE**

FAIL — The diff addresses issue #555 in substance, but it silently changes `Op == "="` handling too and, more importantly, the added test file uses `strings.Contains` without any evidence the `strin

The verification gate passed, but this change may not address the stated objective. Review before merging.

javimosch:
**Automaintainer Review Verdict:** APPROVED — Fix is correct and minimal. All three layers (typecheck, codegen, transform) are consistently patched: `types.go` short-circuits before the undefined-variable check, `transform.go` excludes `_` from declared locals, and `codegen.go` emits `(void)(expr);` to preserve side effects. Three new tests cover all acceptance criteria from #555. The automated intent-check's concern about `strings.Contains` missing an import is incorrect — `strings` is imported on line 4 of assign_undefined_test.go. The handling of both `Op == "="` and `Op == ":="` is intentional and required by AC #1 and #2.

javimosch:

javimosch:
## Architect review — **clean. Merge. First fully green run of the batch.**

Status `published`, not `verify_failed` — the gate actually passed here, and I re-ran it
(`go test -skip 'TestWindowsNetCompiles|TestWindowsBuildsPE'`) to confirm: green.

### Criteria

| # | criterion | verdict |
|---|---|---|
| 1 | `_ = f()` compiles and discards | ✅ |
| 2 | `_ := f()` same, **or** rejected — "pick one and say so" | ✅ same, and SPEC.md says so explicitly |
| 3 | reading `_` back stays an error | ✅ `undefined variable "_"` |
| 4 | `_` consistent as a param / range var, rule documented | ✅ verified below |
| 5 | test alongside the existing assignment tests | ✅ three of them |

### The detail that mattered

`_ = f()` must still **evaluate** the call — a blank destination discards the *value*, not the
side effect. The codegen gets this right (`(void)(expr)`), and I checked it rather than
assuming:

```
var hits = 0
func bump() (v) { hits = hits + 1  v = 42 }
func main() { _ = bump()  _ := bump()  println("side effects ran: " + str(hits)) }

→ side effects ran: 2
```

Both forms ran. An implementation that skipped emission entirely would have printed `0` and
passed every test in this PR, since none of them observe a side effect. Worth adding that
case to the suite.

### Criterion 4, verified empirically

SPEC.md claims *"`_` is write-only in every position"*. Checked:

```
func g(_, b) (v) { v = b }              → accepted
for _, v := range xs { t = t + v }      → 6
x := _                                   → error: undefined variable "_"
```

The claim holds. Good — a spec sentence asserting behaviour in "every position" is exactly
the kind of thing that is usually written from one example.

### One nit

The rule landed in **SPEC.md** but not in **`guide.go`**. `machin guide` is the version-exact
catalog an agent reads at the start of a task — and per this repo's own convention, a
dev-facing behaviour change lands in the guide *and* the docs in the same change. SPEC.md is
the right home for the normative wording; a one-line `gotchas` entry would make it reachable
where agents actually look.

### Recommendation

**Merge** — closes #555. No `[needs work]` prefix on this one, which is itself the correct
signal.

PR #563, opened 2026-08-06
fix(#554): accept field and index lvalues as multi-assign destinations

### LINKED ISSUE #554: multi-assign cannot target a struct field or index (ns.rng, v = f() is a parse error)
## Problem

A multi-value call cannot assign into a struct field. Only plain identifiers are accepted
as destinations.

```
type Rng struct { s int }
func next(r) (nr, v) { nr = r  v = 1 }

func step(s) (ns) {
    ns = s
    ns.rng, v = next(s.rng)     // parse error at the comma
}
```

```
error: parse: unexpected token "," at pos 37
```

The workaround is a temporary per call:

```
r, x := next(s.rng)
ns.rng = r
v = x
```

## Why it matters

MFL structs are value types, so the idiomatic shape is "take the struct, return a modified
copy" — and that means threading sub-structs through multi-value calls constantly. Every
such call costs two extra lines and two throwaway names. In a state-machine-ish program
(a simulation step, a parser, anything carrying an RNG or a cursor) this is the single most
repeated piece of noise.

Indexed destinations have the same limitation: `xs[i], ok = f()`.

## Suggested fix

Accept any assignable lvalue in a multi-assign destination list — identifier, field access
(`a.b.c`), and index (`xs[i]`) — matching what single-assignment already accepts. Evaluate
destinations left to right after the call, as Go does.

## Acceptance criteria

1. `ns.rng, v = next(s.rng)` parses and assigns both.
2. Nested field paths work: `a.b.c, v = f()`.
3. Indexed destinations work: `xs[i], ok = f()`.
4. Mixed with `:=` where at least one name is new, matching current `:=` semantics.
5. Evaluation order is defined and tested (the call once, then the stores).
6. A parser test plus a typecheck test.

## Out of scope

Blank-identifier destinations (`_, v = f()` already works inside a multi-assign); `_` as a
*single* destination is filed separately.

javimosch:

**Focus:** ISSUE FIX OBJECTIVE (GitHub Issue #554: "multi-assign cannot target a struct field or index (ns.rng, v = f() is a parse error)")

ISSUE DESCRIPTION:
## Problem

A multi-value call cannot assign into a struct field. Only plain identifiers are accepted
as destinations.

```
type Rng struct { s int }
func next(r) (nr, v) { nr = r  v = 1 }

func step(s) (ns) {
    ns = s
    ns.rng, v = next(s.rng)     // parse error at the comma
}
```

```
error: parse: unexpected token "," at pos 37
```

The workaround is a temporary per call:

```
r, x := next(s.rng)
ns.rng = r
v = x
```

## Why it matters

MFL structs are value types, so the idiomatic shape is "take the struct, return a modified
copy" — and that means threading sub-structs through multi-value calls constantly. Every
such call costs two extra lines and two throwaway names. In a state-machine-ish program
(a simulation step, a parser, anything carrying an RNG or a cursor) this is the single most
repeated piece of noise.

Indexed destinations have the same limitation: `xs[i], ok = f()`.

## Suggested fix

Accept any assignable lvalue in a multi-assign destination list — identifier, field access
(`a.b.c`), and index (`xs[i]`) — matching what single-assignment already accepts. Evaluate
destinations left to right after the call, as Go does.

## Acceptance criteria

1. `ns.rng, v = next(s.rng)` parses and assigns both.
2. Nested field paths work: `a.b.c, v = f()`.
3. Indexed destinations work: `xs[i], ok = f()`.
4. Mixed with `:=` where at least one name is new, matching current `:=` semantics.
5. Evaluation order is defined and tested (the call once, then the stores).
6. A parser test plus a typecheck test.

## Out of scope

Blank-identifier destinations (`_, v = f()` already works inside a multi-assign); `_` as a
*single* destination is filed separately.

4. COMMIT referencing the issue: fix(#554): <brief description>
5. The PR body MUST include 'Fixes #554' so GitHub auto-closes the issue on merge.

**Branch:** `am/am-84a3be-dkho52so7f4w-478222a5`

## Summary

Fixes #554: multi-assign can now target struct fields and index expressions, not just plain identifiers — `ns.rng, v = next(s.rng)`, `a.b.c, v = f()`, and `xs[i], ok = f()` all parse and assign correctly now. The parser desugars mixed destination lists into a single-call MultiAssign against fresh temps followed by ordered stores, so the RHS call runs exactly once and the checker/codegen/analysis passes needed no changes. Added parser and typecheck tests covering field, nested-field, index, mixed `:=`, and call-once evaluation-order semantics.

**Diff:**
```
multiassign_lvalue_test.go | 144 ++++++++++++++++++++++++++++++++++++++++++
 parser.go                  | 153 ++++++++++++++++++++++++++++++++++-----------
 2 files changed, 259 insertions(+), 38 deletions(-)
```

## Summary by CodeRabbit

* **New Features**
  * Multi-assignment now supports struct fields, nested fields, and indexed destinations.
  * Assignment expressions evaluate the right-hand side once and apply results in left-to-right order.
  * Mixed assignments can combine identifiers with fields or indexes when using `=`.
  * Type checking now validates these complex assignment forms.

* **Bug Fixes**
  * Improved handling of multi-assignment parsing and declaration syntax.

### DIFF
--- a/multiassign_lvalue_test.go
+++ b/multiassign_lvalue_test.go
@@ -1,0 +1,144 @@
+package main
+
+import (
+	"testing"
+)
+
+// TestParseMultiAssignFieldAndIndexDest exercises issue #554: a multi-assign
+// destination list may contain field accesses (including nested paths) and
+// index expressions, not just plain identifiers. The parser desugars this
+// into a temp-destructuring MultiAssign followed by the field/index stores,
+// so the call happens once and stores run left to right.
+func TestParseMultiAssignFieldAndIndexDest(t *testing.T) {
+	prog, err := ParseProgram([]string{
+		`func main() {
+			ns.rng, v = next(s)
+			a.b.c, w = next(s)
+			xs[i], ok = next(s)
+		}`,
+	})
+	if err != nil {
+		t.Fatalf("parse: %v", err)
+	}
+	if prog == nil || len(prog.Funcs) != 1 {
+		t.Fatalf("expected one function, got %v", prog)
+	}
+	body := prog.Funcs[0].Body
+	var multi, fieldStores, indexStores int
+	for _, s := range body {
+		switch s.(type) {
+		case *MultiAssign:
+			multi++
+		case *FieldAssign:
+			fieldStores++
+		case *IndexAssign:
+			indexStores++
+		}
+	}
+	if multi != 3 {
+		t.Fatalf("expected 3 desugared MultiAssign stmts, got %d in %v", multi, body)
+	}
+	if fieldStores != 2 {
+		t.Fatalf("expected 2 FieldAssign stores (ns.rng and a.b.c), got %d", fieldStores)
+	}
+	if indexStores != 1 {
+		t.Fatalf("expected 1 IndexAssign store (xs[i]), got %d", indexStores)
+	}
+}
+
+// TestMultiAssignFieldAndIndexDestTypecheck exercises issue #554 end to end:
+// a multi-value call assigning into a struct field, a nested field path,
+// and an index destination all typecheck once desugared into temp binds
+// plus stores.
+func TestMultiAssignFieldAndIndexDestTypecheck(t *testing.T) {
+	prog, err := ParseProgram([]string{
+		`type Rng struct { s int }`,
+		`type State struct { rng Rng }`,
+		`func next(r) (nr, v) { nr = r  v = 1 }`,
+		`func step(s) (ns) {
+			ns = s
+			v := 0
+			ns.rng, v = next(s.rng)
+			ns.rng.s = ns.rng.s + v
+		}`,
+		`func main() {
+			st := State{rng: Rng{s: 0}}
+			r := step(st)
+			println(r.rng.s)
+		}`,
+	})
+	if err != nil {
+		t.Fatalf("parse: %v", err)
+	}
+	if _, err := Check(prog); err != nil {
+		t.Fatalf("check: %v", err)
+	}
+}
+
+// TestMultiAssignIndexDestTypecheck exercises an index destination (xs[i])
+// mixed with a plain identifier in a multi-assign.
+func TestMultiAssignIndexDestTypecheck(t *testing.T) {
+	prog, err := ParseProgram([]string{
+		`func f() (a, b) { a = 10  b = true }`,
+		`func main() {
+			xs := []int{0, 0, 0}
+			ok := false
+			xs[1], ok = f()
+			println(xs[1])
+			println(ok)
+		}`,
+	})
+	if err != nil {
+		t.Fatalf("parse: %v", err)
+	}
+	if _, err := Check(prog); err != nil {
+		t.Fatalf("check: %v", err)
+	}
+}
+
+// TestMultiAssignEvalOrderCallOnce proves the call happens exactly once
+// (via an observable side effect) before the destinations are stored.
+func TestMultiAssignEvalOrderCallOnce(t *testing.T) {
+	prog, err := ParseProgram([]string{
+		`type Rng struct { s int }`,
+		`var calls = 0`,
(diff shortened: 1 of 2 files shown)


javimosch:
**Automaintainer Review Verdict:** APPROVED — The fix is correct and minimal. `parseStmt()` is cleanly refactored to return `[]Stmt` to support desugaring; mixed lvalue multi-assigns desugar into a single MultiAssign against fresh temps (`_matmpN`) followed by ordered field/index stores, so the RHS call runs exactly once and existing codegen/types/analysis passes are untouched. The `:=` restriction (all-identifier-only) is correctly enforced. Tests cover parse structure, typecheck for field/nested-field/index destinations, call-once semantics, and the existing all-identifier `:=` path. The external Codex comment was a usage-limit notice, not a real review concern. CI is pending but the approach is sound.

javimosch:

javimosch:
## Architect review — **merge**, with one documentation gap worth closing first.

Green, and the full scoped suite passes (427s, zero failures) — which matters here, because
this is the largest change of the batch: `parseStmt` now returns `[]Stmt` and every call site
moved with it (+153/−38 in `parser.go`).

### The approach is the right one

Desugaring in the parser — bind the call's results to temps, then emit the field/index stores
— keeps the type checker and codegen untouched. A multi-assign into `a.b.c` becomes exactly
the code a human writes by hand today, so there is no new lowering path to get wrong.

### Criteria

| # | criterion | verdict |
|---|---|---|
| 1 | `ns.rng, v = next(s.rng)` parses and assigns both | ✅ |
| 2 | nested field paths `a.b.c` | ✅ verified on `b.in.r` |
| 3 | indexed destinations `xs[i], ok = f()` | ✅ |
| 4 | mixed with `:=` where a name is new | ✅ **correctly rejected** — see below |
| 5 | evaluation order defined and tested | ⚠️ tested, **not documented** — see below |
| 6 | parser test + typecheck test | ✅ |

Verified by hand:

```
nested field: s=6 v=50        # b.in.r, v = next(b.in.r)
index: xs1=7 ok=8             # xs[idx()], ok = call()
```

### On criterion 4 — the PR is right and my issue was wrong

I asked for `x.n, fresh := f()` to work. It is rejected:

```
error: parse: cannot declare a field or index with :=
```

**That is the better answer.** `:=` *declares*, and you cannot declare a field — Go rejects
the same program (`non-name x.n on left side of :=`). Good message, too. Ignore that line of
the issue.

### On criterion 5 — order is consistent, but diverges from Go and is undocumented

`TestMultiAssignEvalOrderCallOnce` pins that the call happens **once**, which is the important
safety property. It does not pin the *relative* order of the left-hand index operand against
the call, and that is where MFL now differs from Go. Measured both, same program shape:

```
machin: xs[idx()], ok = call()   → eval order = [call idx ]
go:     xs[idx()], ok = call()   → eval order = [idx call ]
```

Go's spec evaluates *"the operands of index expressions … on the left and the expressions on
the right … in the usual order"* first, then assigns — so `idx()` precedes `call()`. The
desugar here necessarily runs the call first, because the temp bind comes before the stores.

This is **observable only when both the index expression and the call have side effects**, so
it is not a correctness problem and not a blocker. But MFL bills itself as Go-flavored, and
this is precisely the kind of quiet divergence that costs someone an afternoon later.

**Ask:** a sentence in `SPEC.md` (next to the existing multi-assign paragraph) stating the
order — *the right-hand call is evaluated once, then destinations are stored left to right,
so index expressions on the left are evaluated after the call* — and a test that pins the
relative order, not just call-once. Note this PR adds a real language capability and touches
**no** documentation at all; #561, a smaller change, updated SPEC.md.

Minor: the temp is named `_matmp<N>`. Nothing stops a user writing `_matmp1` themselves. A
character illegal in an identifier would make it genuinely unreachable.

### Recommendation

**Merge** — closes #554. The doc sentence can follow, but it should follow.

PR #565, opened 2026-08-06
fix(#557): support slice-range syntax s[lo:hi], s[lo:], s[:hi]

### LINKED ISSUE #557: no slice-range syntax (s[1:]) — every drop-first is a hand-rolled loop
## Problem

There is no slice-range syntax.

```
func main() { s := []int{1,2,3,4}  t := s[1:]  println(str(len(t))) }
```

```
error: parse: expected "]", got ":" at pos 35
```

Every "drop the first element", "take the first N", "the rest of the line" becomes a
hand-rolled loop:

```
func drop_first(xs) (o) {
    o = []int{}
    i := 1
    while i < len(xs) { o = append(o, xs[i])  i = i + 1 }
}
```

## Why it matters

This is the most common operation on a slice after `append` and `len`. Writing it out is
noise, and each hand-rolled copy is a place to get an off-by-one wrong — the demos already
carry several near-identical `drop_first` helpers.

`substr(s, a, b)` already exists for strings, so the concept is in the language; slices just
have no equivalent.

## Suggested fix

Support `s[lo:hi]`, `s[lo:]`, `s[:hi]` on slices, with Go's semantics and bounds rules:
`0 <= lo <= hi <= len(s)`, result length `hi - lo`.

**Aliasing must be decided explicitly and documented.** Go returns a view sharing backing
storage; MFL slices are already described as "reference-ish" for mutation through calls.
Whichever is chosen (view or copy), say so in the `guide.go` entry, because the difference
is observable the first time someone writes through the result. A copy is the safer default
and matches how `substr` behaves on immutable strings; a view is faster. Pick one, document
it, test it.

## Acceptance criteria

1. `s[1:]`, `s[:2]`, `s[1:3]` all compile and produce the right lengths and elements.
2. Out-of-range indices error the same way `s[i]` does today (and under `--safe`, panic).
3. `lo > hi` is an error.
4. Works for `[]int`, `[]string` and `[]<struct>`.
5. Aliasing semantics documented in `guide.go` and pinned by a test that mutates the result
   and checks (or checks the absence of) an effect on the original.

## Out of scope

Three-index slices (`s[a:b:c]`), and slicing strings (use `substr`).

javimosch:

**Focus:** ISSUE FIX OBJECTIVE (GitHub Issue #557: "no slice-range syntax (s[1:]) — every drop-first is a hand-rolled loop")

ISSUE DESCRIPTION:
## Problem

There is no slice-range syntax.

```
func main() { s := []int{1,2,3,4}  t := s[1:]  println(str(len(t))) }
```

```
error: parse: expected "]", got ":" at pos 35
```

Every "drop the first element", "take the first N", "the rest of the line" becomes a
hand-rolled loop:

```
func drop_first(xs) (o) {
    o = []int{}
    i := 1
    while i < len(xs) { o = append(o, xs[i])  i = i + 1 }
}
```

## Why it matters

This is the most common operation on a slice after `append` and `len`. Writing it out is
noise, and each hand-rolled copy is a place to get an off-by-one wrong — the demos already
carry several near-identical `drop_first` helpers.

`substr(s, a, b)` already exists for strings, so the concept is in the language; slices just
have no equivalent.

## Suggested fix

Support `s[lo:hi]`, `s[lo:]`, `s[:hi]` on slices, with Go's semantics and bounds rules:
`0 <= lo <= hi <= len(s)`, result length `hi - lo`.

**Aliasing must be decided explicitly and documented.** Go returns a view sharing backing
storage; MFL slices are already described as "reference-ish" for mutation through calls.
Whichever is chosen (view or copy), say so in the `guide.go` entry, because the difference
is observable the first time someone writes through the result. A copy is the safer default
and matches how `substr` behaves on immutable strings; a view is faster. Pick one, document
it, test it.

## Acceptance criteria

1. `s[1:]`, `s[:2]`, `s[1:3]` all compile and produce the right lengths and elements.
2. Out-of-range indices error the same way `s[i]` does today (and under `--safe`, panic).
3. `lo > hi` is an error.
4. Works for `[]int`, `[]string` and `[]<struct>`.
5. Aliasing semantics documented in `guide.go` and pinned by a test that mutates the result
   and checks (or checks the absence of) an effect on the original.

## Out of scope

Three-index slices (`s[a:b:c]`), and slicing strings (use `substr`).

4. COMMIT referencing the issue: fix(#557): <brief description>
5. The PR body MUST include 'Fixes #557' so GitHub auto-closes the issue on merge.

**Branch:** `am/am-84a3be-dkhpj4594pso-1726c8ad`

**Diff:**
```
arena_check.go     |  6 ++++
 ast.go             |  8 +++++
 codegen.go         | 41 +++++++++++++++++++++++
 guide.go           |  2 +-
 parser.go          | 34 +++++++++++++++++--
 slicerange_test.go | 98 ++++++++++++++++++++++++++++++++++++++++++++++++++++++
 types.go           | 22 ++++++++++++
 7 files changed, 208 insertions(+), 3 deletions(-)
```

### DIFF
--- a/arena_check.go
+++ b/arena_check.go
@@ -79,6 +79,10 @@ func arenaPlaceString(e Expr) (string, bool) {
 		if p, ok := arenaPlaceString(t.X); ok {
 			return p + "#", true
 		}
+	case *SliceExpr:
+		if p, ok := arenaPlaceString(t.X); ok {
+			return p + "#", true
+		}
 	}
 	return "", false
 }
@@ -181,6 +185,8 @@ func detectArenaEscapes(prog *Program, c *Checker) []arenaFinding {
 					return false
 				case *SliceLit:
 					return true
+				case *SliceExpr:
+					return true // mfl_subslice always allocates a fresh copy
 				case *MakeClosure:
 					// a closure created here dangles if it captures a variable that currently
 					// holds arena memory: the capture is read at this make site (inside the
--- a/ast.go
+++ b/ast.go
@@ -127,6 +127,13 @@ type Index struct {
 	Idx Expr
 }
 
+// SliceExpr is a slice-range expression: x[lo:hi]. Lo/Hi are nil when omitted
+// (defaulting to 0 / len(x)). Always produces a fresh copy — see guide.go.
+type SliceExpr struct {
+	X      Expr
+	Lo, Hi Expr
+}
+
 // StructLit constructs a struct value: Point{x: 1, y: 2} or Point{1, 2}.
 // FieldNames is nil for positional literals.
 type StructLit struct {
@@ -181,6 +188,7 @@ func (Binary) node()      {}
 func (Call) node()        {}
 func (SliceLit) node()    {}
 func (Index) node()       {}
+func (SliceExpr) node()   {}
 func (StructLit) node()   {}
 func (FieldAccess) node() {}
 func (FuncLit) node()     {}
--- a/codegen.go
+++ b/codegen.go
@@ -346,6 +346,22 @@ static int64_t mfl_bounds(int64_t i, int64_t n) {
     if (i < 0 || i >= n) { char b[80]; snprintf(b, 80, "index out of range [%lld] with length %lld", (long long)i, (long long)n); mfl_panic(b); }
     return i;
 }
+/* s[lo:hi]: always a fresh copy (never a view sharing s's backing storage) —
+   matches substr's copy semantics on strings and sidesteps cap/aliasing
+   surprises when the result escapes the current block. Bounds are checked
+   unconditionally (like s[i] is under --safe), since an out-of-range slice
+   would otherwise read/copy past the backing array. */
+static mfl_slice mfl_subslice(mfl_slice s, int64_t lo, int64_t hi, int64_t es) {
+    if (lo < 0 || hi < lo || hi > s.len) {
+        char b[96];
+        snprintf(b, 96, "slice bounds out of range [%lld:%lld] with length %lld", (long long)lo, (long long)hi, (long long)s.len);
+        mfl_panic(b);
+    }
+    int64_t n = hi - lo;
+    mfl_slice r = { n ? mfl_alloc(n * es) : NULL, n, n };
+    if (n) memcpy(r.data, (char*)s.data + lo * es, n * es);
+    return r;
+}
 static int64_t mfl_idiv(int64_t a, int64_t b) { if (b == 0) mfl_panic("integer divide by zero"); return a / b; }
 static int64_t mfl_imod(int64_t a, int64_t b) { if (b == 0) mfl_panic("integer modulo by zero"); return a % b; }
 static int64_t mfl_iadd(int64_t a, int64_t b) { int64_t r; if (__builtin_add_overflow(a, b, &r)) mfl_panic("integer overflow (+)"); return r; }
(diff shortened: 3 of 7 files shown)


javimosch:
**Automaintainer Review Verdict:** APPROVED — Clean implementation of slice-range syntax (s[lo:hi] / s[lo:] / s[:hi]). All five acceptance criteria met: correct lengths+elements for int/string/struct slices, copy-not-view semantics pinned by a mutation test, unconditional bounds-check panicking like s[i], and aliasing choice documented in guide.go. The mfl_subslice runtime helper is correct (lo<0 || hi<lo || hi>len check, memcpy into fresh allocation). Parser correctly handles all three forms and the break-out-of-inner-switch pattern is sound Go. Arena escape analysis and type inference updated consistently. No external review concerns to address (both bots hit rate limits).

javimosch:

javimosch:
## Architect review — **merge.** The design decision this issue actually turned on was made, documented, and justified.

Green; full scoped suite passes (544 s, zero failures).

### Criteria

| # | criterion | verdict |
|---|---|---|
| 1 | `s[1:]`, `s[:2]`, `s[1:3]` right lengths and elements | ✅ `4 / 2 / 2`, elems `2,3` |
| 2 | out-of-range errors like `s[i]`; panics under `--safe` | ✅ **unconditionally**, Go-style message |
| 3 | `lo > hi` is an error | ✅ |
| 4 | `[]int`, `[]string`, `[]struct` | ✅ all three |
| 5 | aliasing decided, documented, **and pinned by a mutation test** | ✅ — the important one |

### Criterion 5 — copy, and it is stated where an agent will read it

The issue said: *pick one, document it, test it*, because the difference is observable the
first time someone writes through the result. The PR picks **copy**, and says so in the
`[]T` entry of `guide.go` — the version-exact catalog, not just a comment:

> *"Slice-range `s[lo:hi]` … **ALWAYS returns a fresh COPY**, never a view sharing s's backing
> storage — matches `substr`'s copy semantics on strings and means mutating the result (e.g.
> via append or index-assign) never affects s."*

Verified rather than taken on trust:

```
t := s[1:3]   t[0] = 99
→ s[1] = 2          # 99 here would have meant a view
t = append(t, 77)
→ s len = 5, t len = 3
```

The justification is the right one too: it matches `substr` on strings, so the language is
now internally consistent about what a sub-range means, and it sidesteps the cap/aliasing
questions a view would drag in.

### Bounds

```
s[3:1]  → panic: slice bounds out of range [3:1] with length 3
s[0:9]  → panic: slice bounds out of range [0:9] with length 3
s[-1:]  → panic: slice bounds out of range [-1:3] with length 3
```

Go's exact message shape, and checked **unconditionally** rather than only under `--safe` —
which the guide entry states explicitly. Given the copy is allocated anyway, the check is
free relative to the copy, so this is the right trade.

### The part I did not ask for

`arena_check.go +6`: `mfl_subslice` is registered as an allocating expression, so the escape
analysis knows a sub-range is arena-allocated. That means the new syntax cannot quietly
punch a hole in ARENA001:

```
arena { part := src[1:3]  g = part }
→ warning: [ARENA001] `g` outlives this arena block but is assigned a value
  allocated inside it — it dangles after the block's memory is reclaimed
```

A new allocation site that the escape analysis did not know about would have been a real
soundness gap in machin's headline memory-safety claim. Catching that unprompted is the best
judgment call in the batch.

### One observation, not about this PR

My first attempt at a test for criterion 4 used `[]P{P{1},P{2},P{3}}` and failed — because
this branch is cut from `main` and does not carry #553's non-empty struct literals, which
landed hours earlier in this same batch. That is a concrete instance of
automaintainer-saas-panel#582: sibling issues cannot build on each other. Harmless here;
worth noting as evidence.

### Recommendation

**Merge** — closes #557. Note it took three attempts, both earlier ones lost to harness
failures (an overnight suspend, then an agent waiting on a notification that cannot arrive
in a headless run — panel#592), never to the work itself.

PR #574, opened 2026-08-06
fix(arena): ARENA003 reports a value still live across arena_reset() (#540)

### LINKED ISSUE #540: arena_reset() silently corrupts computed strings held in globals (literals/env/args survive)

javimosch:
Closes #540 — the companion to #539, the same hazard from the other direction.

```
var g_concat = ""
func main() {
    g_concat = "tok-" + env("RT_TOKEN") + "-end"
    println("before: " + g_concat)   // tok-secret123-end
    arena_reset()
    …churn…
    println("after : " + g_concat)   // 1        ← and check said "ok — no errors"
}
```

```
after this PR:
warning: [ARENA003] `g_concat` still holds memory allocated on the arena when
`arena_reset()` frees it — after the reset it reads as garbage (no crash, no
diagnostic). Re-derive it from a literal, `env()` or `args()` after the reset,
or keep it off the arena
```

### It reuses #539's predicate

`arena_reset()` asks the same question ARENA001 does — *was this value allocated here?* — so
the freshness logic is extracted into `arenaFreshness` and shared. That is why the table in the
issue falls out for free rather than being special-cased:

| held across `arena_reset()` | flagged? | why |
|---|---|---|
| string literal | no | static storage |
| `env("X")` / `args()` | no | re-derivation sources, off the arena chain |
| a parameter / another global | no | pre-existing |
| **computed** (concat, append, built slice) | **yes** | allocated here |

A forward scan makes ordering matter: assigning *after* the reset is clean, and clearing a
global to a literal *before* it clears the finding. Eight no-flag cases are pinned by tests.

### One thing the issue got wrong, and I checked rather than assumed

The issue — and the guide table merged in #545 — frame this as a **globals** problem. It is not:

```
func main() { x := "tok-" + env("RT_TOKEN") + "-end"  arena_reset()  churn()  println(x) }
→ after : 1
```

A **local** is corrupted identically. So ARENA003 covers locals too, but only when the local is
actually **read after** the reset — most are dead by then, and a finding nobody can act on is
worse than none. There is a test for each statement shape the liveness scan has to see through
(if / while / range / index / call), because a hazard inside a loop is exactly the one you would
want caught.

### Docs corrected

`SPEC.md` said *"not obvious and not diagnosed"* and `guide.go` said *"no diagnostic"*, both
framing it as globals-only. Both updated. Shipping a fix while the docs still describe the gap
is how the next reader gets misled — and it is the same trap I hit earlier today, when the
bundled gamedev skill still documented the signed-overflow UB that v0.124.0 had just fixed.

### Verification

Full suite green (180s), including `TestTypesCoverageFloor` — which failed first time round at
86.70% against an 87.00% floor and made me write the tests this code owed. Working as intended.

### Honest limit

Unchanged from the rest of the arena analysis: false-negative-permitting. An allocating form not
enumerated in `arenaFreshness` is missed, and liveness for locals is a same-body approximation
(a local escaping into a closure that runs later is not tracked). Every finding is real; not
every hazard is found.

## Summary by CodeRabbit

* **New Features**
  * Added `ARENA003` diagnostics for globals and locals that retain or use arena-allocated memory after `arena_reset()`.
  * Improved detection across nested control flow, reassignment, function calls, and closure environments.
  * Added safeguards to avoid reporting safe non-arena values and duplicate findings.

* **Documentation**
  * Updated `arena_reset()` guidance to cover computed values, slices, bytes, closures, and post-reset local-variable use.
  * Clarified that `ARENA003` findings are advisory.


javimosch:
## Merged with \`--admin\`, bypassing a check that could not run

GitHub Actions is in a **major outage** (githubstatus.com: \`Actions: major_outage\`), and no
workflow run was ever created for this branch — an empty retrigger commit and a close/reopen
both produced nothing. That also explains the eight \`Set up job / Service Unavailable\`
cancellations across today's PRs and the two red \`main\` runs, none of which were real failures.

Owner approved forcing it. Recording what stood in for CI, on the exact commit that merged
(rebased onto current \`main\` first, so this is what lands):

- **Full suite green — 232s, zero failures**, including \`TestTypesCoverageFloor\`.
- Behavioural re-check on the same build:
  - ARENA003 fires on the #540 repro
  - ARENA001 still fires on the #539 repro (shared \`arenaFreshness\` predicate intact)
  - #551's mismatch provenance still names \`charat\`
  - a global assigned a literal across \`arena_reset()\` stays clean (no false positive)
- The sibling change touching these same files (#573) passed real CI earlier today.

If Actions comes back and anything here is red on \`main\`, this is the commit to look at first.

PR #607, opened 2026-08-10
fix(router): unknown paths are distinguishable, and router_init actually resets (#598)

### LINKED ISSUE #598: router: path_index has no not-found signal, and router_init does not reset the route table
Two sharp edges in `framework/router.src`, both found while writing its MFL test
suite (#593). Neither is a crash, and both are pinned by tests as *current*
behaviour rather than silently changed — routing semantics are a product decision.

## 1. `path_index` has no not-found signal

```
func path_index(path) (i) {
    i = 0
    j := 0
    while j < route_n {
        if route_paths[j] == path { i = j }
        j = j + 1
    }
}
```

An unregistered path returns **0**, not -1. So a typo'd or stale deep link
silently lands on whatever route was registered first — usually home — instead of
rendering a 404. The caller cannot tell "matched route 0" from "no idea".

`nav()` is the reason it cannot simply return -1: it does
`navigate(path_index(path))`, and `navigate(-1)` would index `route_paths[-1]`.
Any fix has to decide what an unknown path *means*:

- add an explicit not-found route index the app can register, or
- have `nav()` ignore unknown paths (leaving the current route alone), or
- keep the fallback but document it as intended, and expose a
  `path_known(path)` so an app can 404 deliberately.

The loop also has no early exit and takes the LAST match, so a duplicate
registration silently wins over the first.

## 2. `router_init` does not reset the route table

`route_paths` / `route_n` are package globals that only ever grow. `router_init`
creates a fresh active-route signal but leaves the table alone, so a second
`router_init` does not give a clean slate — indices keep climbing.

A real app calls it once, so this is not a production bug. It bit the test suite
immediately though (every test after the first saw shifted indices), and it makes
the function's name a lie. Either reset the table there or rename it to reflect
that it only initializes the signal.

## Context

Both are covered by `framework/tests/router_test.src`, which asserts the current
behaviour with comments saying it is being pinned, not endorsed. Changing either
means updating those assertions — deliberately, which is the point.

javimosch:
Closes #598. Both edges were found writing router's MFL suite and deliberately left as **product decisions** rather than fixed inside a test. Deciding them now.

## `path_index` returns -1 for an unknown path

It returned **0**, with no not-found signal at all — so a typo'd or stale deep link silently rendered whatever route was registered first (usually home) while the address bar showed the bad URL. No caller could tell *"matched route 0"* from *"no such route"*. They can now, which is what lets an app render a real 404.

`navigate()` is why it couldn't simply return -1: it does `navigate(path_index(path))`, and `route_paths[-1]` reads out of bounds. So **`navigate` now ignores an out-of-range index** — an unrecognised link is inert, current route and address bar unchanged. Least surprising option, and it never crashes. `nav()`, the host entry point for a link click, inherits that.

## First registration wins

`path_index` had no early exit, so it returned the **last** match and a duplicate registration silently shadowed the original. Noted while filing #598; fixed with a `break`.

## `router_init` clears the route table

It only created the signal, so `route_paths`/`route_n` grew for the life of the process and a second `router_init` gave no clean slate — making its name a lie. It bit router's own suite immediately: every test after the first saw shifted indices. Its documented contract is *"call first"*, so resetting is what the name always implied.

## Tests

Three pinned the old behaviour and are **flipped, not deleted** — including one in `reactive_test.go` whose expectation carried a now-contradictory trailing comment. New coverage for each guarantee: -1 vs route 0 distinct, duplicate registration, `navigate` ignoring out-of-range, and `nav()` with an unknown path leaving the route alone.

**32 assertions, router still 9/9.** No other `.src` or example depends on the old semantics (checked). The `machin-web` skill now tells app authors they can 404 on -1, re-synced to `plugins/`.

Full suite green — `ok machin 181.259s`.

### DIFF
--- a/framework/router.src
+++ b/framework/router.src
@@ -6,7 +6,7 @@
 //
 // Usage in a component:
 //   export func start() {
-//       router_init(0)                     // active route is a signal
+//       router_init(0)                     // active route is a signal; also clears routes
 //       route("/")  route("/users")        // register paths, in index order
 //       mount("app", nav_bar() + "<div id=outlet></div>")
 //       outlet("outlet", func() { return page() })   // page() switches on current_route()
@@ -31,8 +31,13 @@ var route_now = 0           // signal id: the active route index
 var route_paths = []string{}
 var route_n = 0
 
-// router_init creates the active-route signal (call first, with the initial index).
+// router_init creates the active-route signal AND clears the route table (call
+// first, with the initial index). It resets: previously it only made the signal,
+// so a second router_init left the old paths in place and indices kept climbing —
+// which made its name a lie and broke any test or app that set up twice (#598).
 func router_init(initial) {
+    route_paths = []string{}
+    route_n = 0
     route_now = signal(initial)
 }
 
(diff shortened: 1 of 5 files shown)


javimosch:
Superseded by #606, which landed the same two fixes concurrently (path_index → -1, router_init resets).

Closing rather than force-merging a duplicate. One genuine gap #606 left is being opened as a focused follow-up: `navigate` guards `i < 0` but not the upper bound, so `navigate(99)` still reads past the table — reproduced on main:

```
panic: index out of range [99] with length 2
```

Plus the tests for it and the machin-web skill note that app authors can now 404 on -1.

Issue #578, opened 2026-08-07
append's growth path is the entire sieve benchmark gap (not slice indexing) — and the obvious realloc fix is unsound


javimosch:
`bench/native-speed` has always shown machin trailing Rust and Zig on the sieve
kernel by ~1.4x, and the benchmark README explained it as:

> On the **array-heavy sieve** machin trails by ~1.4× — its slice indexing/layout
> is less optimal than a Rust `Vec` or a Zig slice.

**That diagnosis is wrong.** Phase-timing the kernel shows slice indexing is not
the problem at all — the sieve loop ties Rust exactly:

| phase | machin | Rust |
|---|--:|--:|
| build the 10M array by `append` | **70–83 ms** | **27–29 ms** |
| the sieve loop itself | 110 ms | 111 ms |
| the count loop | 5–7 ms | 3 ms |

The generated C for the hot loop is already what you would write by hand — direct
pointer indexing, no bounds check, no indirection:

```c
while ((v_m <= v_n)) { ((int64_t*)(v_sieve).data)[v_m] = 0LL; v_m = (v_m + v_p); }
```

The entire gap is `append`'s growth path.

## Mechanism

`mfl_append` doubles capacity through `mfl_realloc`, and `mfl_realloc` can only
ever *copy*:

```c
static void* mfl_realloc(void* old, size_t sz) {
    void* p = mfl_alloc(sz);                 /* a brand-new malloc'd arena block */
    if (old) { size_t o = ((mfl_blk*)old - 1)->size; memcpy(p, old, o < sz ? o : sz); }
    return p; /* old reclaimed with its arena — never freed individually */
}
```

So growing to 10M elements walks ~21 doublings, each one allocating a fresh block
and memcpy'ing, and **never releasing the previous buffer** (an arena frees nothing
mid-life). The process therefore touches roughly 2x the final array in freshly
faulted pages. `Vec::push` hands the block to `realloc`, and glibc extends a large
block in place via `mremap` — no copy, no new pages.

## The obvious fix is unsound — do not just do it

When the block being grown is the arena's most recent allocation, nothing is
layered on top of it in the block list, so it looks safe to hand it straight to
`realloc()` and let glibc `mremap` it. I implemented and then rejected this,
because MFL slices **share backing storage**:

```
func mutate(s) { s[0] = 77 }

a := []int{1,2,3}
b := a            // b.data == a.data  (verified: mutating a changes b)
mutate(c)         // params share too  (verified)
```

[...]

PR #590, opened 2026-08-09
[needs work] fix(windows): enable raw_mode/read_key on the Windows target via conio

javimosch:
Automated maintenance run by automaintainer.

**Focus:** == ASSIGNED OBJECTIVE ==
Fix open GitHub issues. Small PRs, conventional commits.

----
OPEN PR AWARENESS (secondary — do not replace the ASSIGNED OBJECTIVE):
These open pull requests are already open and awaiting review. Do NOT start UNRELATED work on the files they touch. If your ASSIGNED OBJECTIVE requires editing one of those files, complete the objective anyway. Never abandon the objective to pick a different GitHub issue just to avoid overlap.

- PR #585 (am/am-7b5889-dkjnye272rgj-6b99e560): [needs work] fix(windows): bundle SQLite amalgamation for --target windows
    touches: build.go, codegen.go, guide.go, windows_target_test.go
- PR #584 (am/am-7b5889-dkjfvfst437e-d72f729d): [needs work] feat(slice): add make([]T, n) and make([]T, len, cap)
    touches: README.md, SPEC.md, ast.go, bench/native-speed/README.md, bench/native-speed/machin/sieve.src, codegen.go, docs/LANGUAGE.md, examples/complex/slice_prealloc.mfl, guide.go, makeslice_test.go, parser.go, parsetest.go (+4 more)
- PR #583 (am/am-7b5889-dkj8t8dwx2ph-cec70d48): feat(slice): make([]T, n) and make([]T, len, cap) — fixes #578
    touches: README.md, SPEC.md, ast.go, bench/native-speed/machin/sieve.src, codegen.go, docs/LANGUAGE.md, guide.go, makeslice_test.go, parser.go, parsetest.go, render.go, transform.go (+1 more)
- PR #582 (am/am-7b5889-dkj1a1zi5nb0-c57763ed): feat: sort / sort_by builtins (#580)
    touches: SPEC.md, codegen.go, docs/LANGUAGE.md, guide.go, sort_test.go, types.go
- PR #577 (am/am-7b5889-dkidr8jtcv78-d28a4fc8): [needs work] fix(windows): enable XEdDSA (xeddsa_sign/verify) on the Windows target
    touches: build.go, codegen.go, guide.go, main.go, windows_target_test.go

**Branch:** `am/am-7b5889-dkjzkfzyn4r9-799cb52e`

> ⚠️ **Verification failed** — this PR is a draft. Last output:
```
[...]

PR #585, opened 2026-08-08
[needs work] fix(windows): bundle SQLite amalgamation for --target windows

javimosch:
Automated maintenance run by automaintainer.

**Focus:** == ASSIGNED OBJECTIVE ==
Fix open GitHub issues. Small PRs, conventional commits.

----
OPEN PR AWARENESS (secondary — do not replace the ASSIGNED OBJECTIVE):
These open pull requests are already open and awaiting review. Do NOT start UNRELATED work on the files they touch. If your ASSIGNED OBJECTIVE requires editing one of those files, complete the objective anyway. Never abandon the objective to pick a different GitHub issue just to avoid overlap.

- PR #584 (am/am-7b5889-dkjfvfst437e-d72f729d): [needs work] feat(slice): add make([]T, n) and make([]T, len, cap)
    touches: README.md, SPEC.md, ast.go, bench/native-speed/README.md, bench/native-speed/machin/sieve.src, codegen.go, docs/LANGUAGE.md, examples/complex/slice_prealloc.mfl, guide.go, makeslice_test.go, parser.go, parsetest.go (+4 more)
- PR #583 (am/am-7b5889-dkj8t8dwx2ph-cec70d48): feat(slice): make([]T, n) and make([]T, len, cap) — fixes #578
    touches: README.md, SPEC.md, ast.go, bench/native-speed/machin/sieve.src, codegen.go, docs/LANGUAGE.md, guide.go, makeslice_test.go, parser.go, parsetest.go, render.go, transform.go (+1 more)
- PR #582 (am/am-7b5889-dkj1a1zi5nb0-c57763ed): feat: sort / sort_by builtins (#580)
    touches: SPEC.md, codegen.go, docs/LANGUAGE.md, guide.go, sort_test.go, types.go
- PR #577 (am/am-7b5889-dkidr8jtcv78-d28a4fc8): [needs work] fix(windows): enable XEdDSA (xeddsa_sign/verify) on the Windows target
    touches: build.go, codegen.go, guide.go, main.go, windows_target_test.go

**Branch:** `am/am-7b5889-dkjnye272rgj-6b99e560`

> ⚠️ **Verification failed** — this PR is a draft. Last output:
```
	/usr/lib/go-1.22/src/os/exec/exec.go:991 +0xb6 fp=0xc0001ffe78 sp=0xc0001ffe28 pc=0x516a96
machin.RunCaptured(0xc000294ea0)
[...]

PR #584, opened 2026-08-08
[needs work] feat(slice): add make([]T, n) and make([]T, len, cap)

javimosch:
Automated maintenance run by automaintainer.

**Focus:** == ASSIGNED OBJECTIVE ==
Fix open GitHub issues. Small PRs, conventional commits.

----
OPEN PR AWARENESS (secondary — do not replace the ASSIGNED OBJECTIVE):
These open pull requests are already open and awaiting review. Do NOT start UNRELATED work on the files they touch. If your ASSIGNED OBJECTIVE requires editing one of those files, complete the objective anyway. Never abandon the objective to pick a different GitHub issue just to avoid overlap.

- PR #583 (am/am-7b5889-dkj8t8dwx2ph-cec70d48): feat(slice): make([]T, n) and make([]T, len, cap) — fixes #578
    touches: README.md, SPEC.md, ast.go, bench/native-speed/machin/sieve.src, codegen.go, docs/LANGUAGE.md, guide.go, makeslice_test.go, parser.go, parsetest.go, render.go, transform.go (+1 more)
- PR #582 (am/am-7b5889-dkj1a1zi5nb0-c57763ed): feat: sort / sort_by builtins (#580)
    touches: SPEC.md, codegen.go, docs/LANGUAGE.md, guide.go, sort_test.go, types.go
- PR #577 (am/am-7b5889-dkidr8jtcv78-d28a4fc8): [needs work] fix(windows): enable XEdDSA (xeddsa_sign/verify) on the Windows target
    touches: build.go, codegen.go, guide.go, main.go, windows_target_test.go

**Branch:** `am/am-7b5889-dkjfvfst437e-d72f729d`

> ⚠️ **Verification failed** — this PR is a draft. Last output:
```
created by testing.(*T).Run in goroutine 1
	/usr/lib/go-1.22/src/testing/testing.go:1742 +0x390

goroutine 2146 [IO wait, 118 minutes]:
internal/poll.runtime_pollWait(0x77c7219a5c08, 0x72)
	/usr/lib/go-1.22/src/runtime/netpoll.go:345 +0x85
internal/poll.(*pollDesc).wait(0xc002770960?, 0xc002864000?, 0x1)
	/usr/lib/go-1.22/src/internal/poll/fd_poll_runtime.go:84 +0x27
internal/poll.(*pollDesc).waitRead(...)
	/usr/lib/go-1.22/src/internal/poll/fd_poll_runtime.go:89
[...]

PR #612, opened 2026-08-10
feat(windows): regex works on the windows target, verified by running it (#517)

javimosch:

**Focus:** == ASSIGNED OBJECTIVE ==
Fix open GitHub issues. Small PRs, conventional commits.

----
OPEN PR AWARENESS (secondary — do not replace the ASSIGNED OBJECTIVE):
These open pull requests are already open and awaiting review. Do NOT start UNRELATED work on the files they touch. If your ASSIGNED OBJECTIVE requires editing one of those files, complete the objective anyway. Never abandon the objective to pick a different GitHub issue just to avoid overlap.

- PR #590 (am/am-7b5889-dkjzkfzyn4r9-799cb52e): [needs work] fix(windows): enable raw_mode/read_key on the Windows target via conio
    touches: SPEC.md, build.go, codegen.go, guide.go, windows_target_test.go
- PR #585 (am/am-7b5889-dkjnye272rgj-6b99e560): [needs work] fix(windows): bundle SQLite amalgamation for --target windows
    touches: build.go, codegen.go, guide.go, windows_target_test.go
- PR #584 (am/am-7b5889-dkjfvfst437e-d72f729d): [needs work] feat(slice): add make([]T, n) and make([]T, len, cap)
    touches: README.md, SPEC.md, ast.go, bench/native-speed/README.md, bench/native-speed/machin/sieve.src, codegen.go, docs/LANGUAGE.md, examples/complex/slice_prealloc.mfl, guide.go, makeslice_test.go, parser.go, parsetest.go (+4 more)
- PR #583 (am/am-7b5889-dkj8t8dwx2ph-cec70d48): feat(slice): make([]T, n) and make([]T, len, cap) — fixes #578
    touches: README.md, SPEC.md, ast.go, bench/native-speed/machin/sieve.src, codegen.go, docs/LANGUAGE.md, guide.go, makeslice_test.go, parser.go, parsetest.go, render.go, transform.go (+1 more)
- PR #582 (am/am-7b5889-dkj1a1zi5nb0-c57763ed): feat: sort / sort_by builtins (#580)
    touches: SPEC.md, codegen.go, docs/LANGUAGE.md, guide.go, sort_test.go, types.go
- PR #577 (am/am-7b5889-dkidr8jtcv78-d28a4fc8): [needs work] fix(windows): enable XEdDSA (xeddsa_sign/verify) on the Windows target
    touches: build.go, codegen.go, guide.go, main.go, windows_target_test.go

**Branch:** `am/am-7b5889-dklkh77633dn-8db405bc`

**Diff:**
```
build.go               |   39 +-
 codegen.go             |  124 +++-
 guide.go               |    2 +-
 vendor/regex/README.md |   11 +
 vendor/regex/remimu.h  | 1509 ++++++++++++++++++++++++++++++++++++++++++++++++
 windows_target_test.go |   48 +-
 6 files changed, 1719 insertions(+), 14 deletions(-)
```

## Summary by CodeRabbit

* **New Features**
  * Windows-targeted builds now support SQLite and POSIX-compatible regular expressions.
  * Regular expression matching, searching, grouping, and replacement are available on Windows.
  * Invalid patterns and matching limits are handled safely.

* **Documentation**
  * Updated Windows build guidance and capability descriptions to reflect SQLite and regex support.
  * Added documentation and licensing details for the bundled regex engine.

* **Tests**
  * Added coverage for Windows regex compilation and end-to-end executable builds.


javimosch:
Heads-up: #611 (zlib for the windows target) just merged and touches the same four files as this PR — `build.go`, `codegen.go`, `guide.go`, `windows_target_test.go`. This will need a rebase.

Two things from the zlib side that may save you time:

- The windows preflight list in `windowsUnsupported` shrank again, so the rejection cases in `windows_target_test.go` have moved. zlib is gone from it; regex is still there and is presumably what you remove.
- `unpackZlibSources` in `build.go` sits right next to `unpackSqliteAmalgamation`, and the windows build now has a bundling block per vendored library. If `remimu.h` is header-only it is simpler than either — just an `-I` — but the same `usesX := strings.Contains(csrc, ...)` + block shape should slot in cleanly.

Also worth adding if it is not there: CI now runs the cross-compiled `.exe` on a `windows-latest` job rather than only compiling it (`windows-run`), because this target was broken across two releases while CI happily built it (#549). Both SQLite and zlib assert real output there; a regex one would fit the same pattern.

javimosch:
🤖 Escalated to CEO: https://github.com/javimosch/am-fleet/issues/53

The rebaser could not resolve conflicts automatically. Reply on the issue above to unblock this PR.

javimosch:
Reviewed rather than merged — there is one blocking concern, and it is specific.

## Remimu and POSIX ERE do not agree on alternation

machin's native regex is POSIX `<regex.h>`, which is **leftmost-longest**. Verified on `main` just now:

```
regex_find("a|ab",     "ab")  -> "ab"
regex_find("x|xy|xyz", "xyz") -> "xyz"
```

Remimu is a **backtracking** engine — its own docs mention possessive and atomic groups, which are PCRE constructs, and backtracking engines are leftmost-**first**. The same two calls should return `"a"` and `"x"` under it.

If that is right, this PR makes **the same machin program return different results on Linux and Windows**, silently. No error, no warning — `regex_find` just answers differently. That is the failure mode #549 was (a Windows claim nobody executed), one level deeper.

## What would make it landable

Either would do:

1. **A differential test.** Cross-compile a program exercising alternation, greediness, nested quantifiers and anchors; run it on the `windows-latest` job (that infrastructure exists now — see `windows-run`, which asserts real output for SQLite, zlib and a goroutine fan-in) and assert it matches the native run **byte for byte**. If they agree, the concern is dead and the PR is strictly good.
2. **Document the divergence honestly** — in `machin guide`, in the builtin's description, and in the #517 notes: "regex on the windows target is a backtracking engine; alternation is leftmost-first, not POSIX leftmost-longest." Then it is a known, stated limitation instead of a trap.

I would push for (1), because a cross-platform semantic difference in a *builtin* is the kind of thing that costs someone a day and never gets traced back to the compiler.

## Also, this needs a rebase

It conflicts with `main` (`MERGEABLE: CONFLICTING`). Since it was opened, zlib (#611) and select (#619) both landed and touched the same four files. The preflight list in `windowsUnsupported` has shrunk twice — `select` and `zlib` are both gone from it now, so the guide string this PR edits has moved.

The vendoring pattern it uses is right and matches what zlib did (`vendor/regex/README.md` with the licence, bundled at build time, no external dependency). It is the semantics, not the packaging, that needs the extra work.

javimosch:
Measured this rather than argue it, and it does not hold up. Harness + full numbers in #634.

**13 of 50 divergences** against POSIX `<regex.h>`, over the exact operations `regex_*` exposes, in three classes:

| class | example | native (POSIX) | this PR on Windows |
|---|---|---|---|
| leftmost-**longest** vs leftmost-**first** | `b\|bc` on `abcd` | `bc` | `b` |
| POSIX bracket classes unsupported | `[[:digit:]]+` on `x42y` | `42` | **no match** |
| PCRE escapes POSIX takes literally | `\d+` on `x42y` | **no match** | `42` |

Seven are the classic split — POSIX takes the longest alternative, a backtracker the first that succeeds. `(foo\|foobar)` on `foobar` gives `foobar` natively and `foo` here.

The second class is worse than a different answer: Remimu is PCRE-flavoured (`\d`, `\s`, `\w`), so a program filtering on `[[:digit:]]+` finds **nothing at all** on Windows. I checked for a parse flag enabling POSIX classes — there isn't one, `REMIMU_FLAG_DOT_NO_NEWLINES` is the only flag.

The third runs the other way: a pattern that works **only** on Windows and breaks when the same program is built natively.

None of this is a knock on Remimu, which is a tidy single-header engine. It is the wrong *shape* for this job: the requirement here is not "a regex engine" but "the same answers `<regex.h>` gives", because the alternative is one MFL program returning different results per platform with no oracle to catch it — the exact failure class #625 was about.

**Closing.** Keeping the honest compile-time rejection is better than silently different behaviour. #634 lands the harness so the bar is executable: a candidate must be **leftmost-longest** and support **POSIX bracket expressions**, at zero divergences. A DFA/TNFA engine (musl's `regcomp`/`regexec`, TRE) meets that; the cheapest sound option is one engine on *every* target so the two cannot drift — which must clear the same harness, since it changes native behaviour too.

The vendored `remimu.h` and the `--target windows` wiring in this branch stay available in git if a future attempt wants the plumbing.

